Issue #65, opened 2016-08-28
Gradle plugin relies on Gradle Groovy plugin


shanman190:
I know that this project is still in the very early stages, but I have a lot of interest in the idea, so I'm playing with it VERY early.

The current version of the Gradle plugin requires that the Groovy plugin be applied. It's fairly likely that this library could be included in a project that doesn't have the Groovy plugin already applied as that project may not have Groovy source code. 

The current behavior is that the GenerateServerTestsTask results in an exception during execution using either the milestone release or the current snapshot available.

There are three options to handle work around this:

1) Apply the Groovy plugin within the setup of the Contracts plugin,
2) Log a warning stating that the plugin was not configured and short circuit the plugin application process, or
3) Document that the Groovy plugin is a requirement

marcingrzejszczak:
Hello!! Thanks for the feedback :)

Btw this project is built on top of another project called Accurest which was GA already a couple if months ago so actually Spring Cloud Contract is not at that early stage :)

Anyways you're right! We should automatically add the plugin if it's not there. I'll definitely have to fix this :)

PR #66, opened 2016-08-28
Updated plugin to explicitly include Gradle Groovy plugin

shanman190:
Updated plugin to explicitly include Gradle Groovy plugin, so that this plugin can properly evaluate when included in a Gradle project that does not currently apply the Groovy plugin. Additionally, added unit tests to verify functionality provided by the plugin. Fixes gh-65


marcingrzejszczak:
This is super awesome @shanman190 !! Thanks a lot :D

marcingrzejszczak:
I've twitted about you :P https://twitter.com/MGrzejszczak/status/770146248179453952

PR #106, opened 2016-10-12
add spring-amqp/spring-rabbit support for message contracts

mduesterhoeft:
Add spring-rabbit integration for messaging contracts working in local mode without a connected bus.


mduesterhoeft:
This PR is work-progress. There are test like `spring-cloud-contract-stub-runner-integration` missing to test the subscriber side.

I add this in an early state to get feedback if the directions I took make sense and would find the approval of the maintainers.

Looking forward to your feedback.

pom.xml
```diff
@@ -105,6 +105,16 @@
 				<version>0.5.0</version>
 			</dependency>
 			<dependency>
+				<groupId>org.springframework.amqp</groupId>
+				<artifactId>spring-rabbit</artifactId>
+				<version>1.6.2.RELEASE</version>
+			</dependency>
+			<dependency>
```
marcingrzejszczak: I think that Mockito comes with Spring Boot Test doesn't it? If that's the case you don't have to explicitly provide it here

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/ContractVerifierAmqpAutoConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+import org.springframework.cloud.contract.verifier.messaging.internal.ContractVerifierMessage;
+import org.springframework.cloud.contract.verifier.messaging.internal.ContractVerifierMessaging;
+import org.springframework.cloud.contract.verifier.messaging.stream.ContractVerifierStreamAutoConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Configuration setting up {@link MessageVerifier} for use with plain spring-rabbit/spring-amqp
+ *
+ * @author Mathias Düsterhöft
```
marcingrzejszczak: `@since 1.0.1`

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/ContractVerifierAmqpAutoConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+@Configuration
+@ConditionalOnClass({Message.class, Mockito.class})
```
marcingrzejszczak: I don't like this approach with Mockito (which doesn't mean that it's wrong). What about we create our own extension of `RabbitTemplate`? Sth like this (pseudocode) - note that I don't even know if it will work :P 

```
class OurRabbitTemplate extends RabbitTemplate {

List<Message> messages = new ArrayList<>();

@Override
send(...,Message m, ...) {
messages.add(m);
}


}

``` 

Other than that (once again, I'm not saying that this approach is wrong - I'm just thinking out loud about sth that is nor relying on Mockito) it seems ok :)

marcingrzejszczak: I think we should have `RabbitTemplate` here

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/SpringAmqpStubMessages.java
```diff
@@ -0,0 +1,108 @@
...
+	private static final Logger log = LoggerFactory.getLogger(SpringAmqpStubMessages.class);
+
+	private final RabbitTemplate rabbitTemplate;
+
+	private final MessageListenerAdapter messageListenerAdapter;
+
+	@Autowired
+	public SpringAmqpStubMessages(RabbitTemplate rabbitTemplate, MessageListenerAdapter messageListenerAdapter) {
+		Assert.notNull(rabbitTemplate);
+		Assert.isTrue(Mockito.mockingDetails(rabbitTemplate).isSpy());
```
marcingrzejszczak: Why do we need this to be a spy?

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/ContractVerifierAmqpAutoConfiguration.java
mduesterhoeft: I also thought about this - the downside is that the client would have to configure the OurRabbitTemplate - this is very intrusive - in order to use the message contract with rabbit mq you need to configure another rabbit template.

The spy-approach is less intrusive.

marcingrzejszczak: makes sense 👍

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/SpringAmqpStubMessages.java
marcingrzejszczak: 🆗

pom.xml
marcingrzejszczak: Ah yeah, right. :+1:

mduesterhoeft:
@marcingrzejszczak please check - I added the missing tests. Is it ready to merge?

spring-cloud-contract-verifier/pom.xml
```diff
@@ -37,6 +37,16 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.amqp</groupId>
+			<artifactId>spring-rabbit</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<!--<dependency>-->
```
marcingrzejszczak: Let's maybe remove this ?

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/ContractVerifierAmqpAutoConfiguration.java
marcingrzejszczak: Can you add this? ^^

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/SpringAmqpStubMessages.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * {@link MessageVerifier} implementation to integrate with plain spring-amqp/spring-rabbit.
+ * It is meant to be used without interacting with a running bus.
+ *
+ * It relies on the RabbitTemplate to be a spy to be able to capture send messages.
+ *
+ * Messages are not sent to the bus - but are handed over to the {@link MessageListenerAdapter} which
+ * allows us to test the full deserialization and listener invocation.
+ *
+ * @author Mathias Düsterhöft
```
marcingrzejszczak: missing `@since 1.0.1`

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/ContractVerifierAmqpAutoConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Configuration setting up {@link MessageVerifier} for use with plain spring-rabbit/spring-amqp
+ *
+ * @author Mathias Düsterhöft
+ */
+@Configuration
+@ConditionalOnClass({Message.class, RabbitTemplate.class})
```
marcingrzejszczak: I think that `Mockito` should remain here. There can be someone using EasyMock or sth that would like to use AMQP support. If one doesn't have Mockito it will blow up at runtime, won't it?

spring-cloud-contract-verifier/src/test/groovy/org/springframework/cloud/contract/verifier/messaging/amqp/SpringAmqpStubMessagesSpec.groovy
```diff
@@ -0,0 +1,39 @@
...
+            SpringAmqpStubMessages messageVerifier = new SpringAmqpStubMessages(rabbitTemplate, messageListenerAdapter)
+        when:
+            messageVerifier.send(payload,
+                    ImmutableMap.builder()
+                            .put(DEFAULT_CLASSID_FIELD_NAME, "org.example.Some")
+                            .put("contentType", CONTENT_TYPE_JSON)
+                            .build(),
+                    "test-exchange")
+        then:
+            then(messageListenerAdapter).should().onMessage(messageArgumentCaptor.capture())
```
marcingrzejszczak: you could have used Spock's mocks but that's not a problem :)

tests/samples-messaging-amqp/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-actuator</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-amqp</artifactId>
+		</dependency>
+		<dependency>
```
marcingrzejszczak: Would we have to require from users to explicitly provide `mockito` in compile scope? That might be bizarre. I think that our starter should provide `mockito` in compile scope, whereas  the user uses our starter in test scope. Or am I missing sth?

tests/samples-messaging-amqp/src/main/java/com/example/AmqpMessagingApplication.java
```diff
@@ -0,0 +1,68 @@
...
+		return rabbitTemplate;
+	}
+
+	@Bean
+	public Exchange testExchange() {
+		return new TopicExchange("test-exchange");
+	}
+
+	@Bean
+	public ConnectionFactory connectionFactory() {
```
marcingrzejszczak: If this is the typical setup I think we should mention this in the docs

mduesterhoeft:
Thanks for the comments - they should be all applied now - I will add the documentation.

mduesterhoeft:
Sorry for the delay - I added a basic documentation. I also realized that the current solution does not support `@RabbitListener`-based listener endpoints. I will try to add this later in another PR.

tests/spring-cloud-contract-stub-runner-amqp/README.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+The integration tries to work standalone, that is without interaction with a running RabbitMQ server. It expects a `RabbitTemplate` on the application context and makes it a spring boot test `@SpyBean`.
+Thus it can use the mockito spy to intercept messages sent by the application.
+
+IMPORTANT: In the current status of the implementation tries to find a `MessageListenerAdapter` on the application context to send messages into the application, because we want to avoid interacting with a running bus.
+Using annotation driven listener endpoints (e.g. `@RabbitListener` annotated listener methods) is currently not supported.
+
+==== Adding it to the project
+
+It's enough to have both Apache Camel and Spring Cloud Contract Stub Runner on classpath.
```
marcingrzejszczak: `Apache Camel` -> `Spring AMQP` ?

tests/spring-cloud-contract-stub-runner-amqp/src/test/groovy/org/springframework/cloud/contract/stubrunner/messaging/amqp/AmqpStubRunnerSpec.groovy
```diff
@@ -0,0 +1,68 @@
...
+            then(messageSubscriber).should().handleMessage(personArgumentCaptor.capture())
+            personArgumentCaptor.value.name != null
+    }
+
+    @Configuration
+    static class TestConfig {
+        /**
+         * Do not connect to a real bus - so we mock the connection factory
+         */
+        @Bean
```
marcingrzejszczak: I'd put this part also in the docs. To show the connection factory should be configured. Come to think of it - maybe we should include it as part of the AutoConfig ?

mduesterhoeft: It should definitely be mentioned in the docs. I would shy away from adding it to the AutoConfiguration because a client might actually want to run against a running bus - Or are you wanting to test in isolation by design in spring-cloud-contract.

Your decision.

marcingrzejszczak: Yeah I think that it makes sense. Take a look at what we do with Service Discovery. Example for Eureka - https://github.com/spring-cloud/spring-cloud-contract/blob/master/spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/spring/cloud/eureka/StubRunnerSpringCloudEurekaAutoConfiguration.java#L49 . By default it is *disabled*. But Autoconfiguration to allow people to register stubs in Eureka is there. https://github.com/spring-cloud/spring-cloud-contract/blob/master/spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/spring/cloud/ConditionalOnStubbedDiscoveryDisabled.java#L37 . WDYT? Stub it out by default but give an option to turn it on.

marcingrzejszczak: The mocked connection factory is *enabled by default* - we don't want to use any real communication by default. You can *turn it off* by property if you want to use a real one.

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/ContractVerifierAmqpAutoConfiguration.java
marcingrzejszczak: Meh now it should be `1.0.2` :/

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/SpringAmqpStubMessages.java
marcingrzejszczak: Meh now it should be `1.0.2` :/

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/RabbitMockConnectionFactoryAutoConfiguration.java
```diff
@@ -23,7 +23,7 @@
  * @since 1.0.2
  */
 @Configuration
-@ConditionalOnProperty(value = "verifier.amqp.mockConnection", havingValue = "true")
+@ConditionalOnProperty(value = "verifier.amqp.mockConnection", havingValue = "true", matchIfMissing = true)
```
marcingrzejszczak: :+1:

tests/spring-cloud-contract-stub-runner-amqp/README.adoc
```diff
@@ -0,0 +1,110 @@
...
+├── META-INF
+│   └── MANIFEST.MF
+└── contracts
+    └── shouldProduceValidPersonData.groovy
+----
+
+Let's consider the following contract:
+
+[source,groovy]
+----
```
marcingrzejszczak: From what I see this is the last thing to fix. Can you embed this contract in some test class and reference it with `include` directive? That's because if we ever refactor the structure then this part of the doc will get updated too

marcingrzejszczak:
@mduesterhoeft  Oops some tests failed with `java.lang.NoClassDefFoundError: org/springframework/amqp/rabbit/connection/ConnectionFactory`. Most likely we need on the Spring AMQP autoconfig the conditional on `org/springframework/amqp/rabbit/connection/ConnectionFactory`

mduesterhoeft:
@marcingrzejszczak exactly - the conditions on the autoConfigurations for the mock ConnectionFactory are not sufficient...

mduesterhoeft:
@marcingrzejszczak should be fixed now

marcingrzejszczak:
Congrats!

marcingrzejszczak:
It failed on our Jenkins - I'll check it out later

```
Error Message

No bean named 'contract-test.exchange' is defined
Stacktrace

org.springframework.beans.factory.NoSuchBeanDefinitionException: No bean named 'contract-test.exchange' is defined
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBeanDefinition(DefaultListableBeanFactory.java:677)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getMergedLocalBeanDefinition(AbstractBeanFactory.java:1180)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:284)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:202)
	at org.springframework.context.support.AbstractApplicationContext.getBean(AbstractApplicationContext.java:1082)
	at org.springframework.cloud.contract.verifier.messaging.integration.SpringIntegrationStubMessages.send(SpringIntegrationStubMessages.java:58)
	at org.springframework.cloud.contract.verifier.messaging.integration.SpringIntegrationStubMessages.send(SpringIntegrationStubMessages.java:52)
	at org.springframework.cloud.contract.stubrunner.StubRunnerExecutor.sendMessageIfApplicable(StubRunnerExecutor.java:214)
	at org.springframework.cloud.contract.stubrunner.StubRunnerExecutor.triggerForDsls(StubRunnerExecutor.java:176)
	at org.springframework.cloud.contract.stubrunner.StubRunnerExecutor.trigger(StubRunnerExecutor.java:162)
	at org.springframework.cloud.contract.stubrunner.StubRunner.trigger(StubRunner.java:98)
	at org.springframework.cloud.contract.stubrunner.BatchStubRunner.trigger(BatchStubRunner.java:131)
	at org.springframework.cloud.contract.stubrunner.messaging.amqp.AmqpStubRunnerSpec.should trigger stub amqp message(AmqpStubRunnerSpec.groovy:61)
```

marcingrzejszczak:
BTW did you run `./mvnw clean install -Pdocs,integration`  locally ?

marcingrzejszczak:
Ah, you've put an old version `1.0.1` in the test pom. Once I fixed it seems to work (at least on my machine - I'm waiting for Jenkins)

mduesterhoeft:
Ah OK - in the meantime its 1.0.2?!

marcingrzejszczak:
I mean we've just released 1.0.1.RELEASE so in the Pom it should be 1.0.2.BUILD-SNAPSHOT

mduesterhoeft:
should I fix that?

marcingrzejszczak:
Nope - I already have!

PR #125, opened 2016-10-28
Fixes #124 - When specifying 'workOffline=false', the stubrunner will…

### LINKED ISSUE #124: StubRunnerConfiguration defaults to offline mode when repository root is not specified
I think the `workOffline` property should take precedence, regardless of whether or not there's a `repositoryRoot` or not.  In otherword I would expect:

| Valid Repository Root Specified  | Work Offline | Outcome |
| ------------- | ------------- | ------------- |
| True  | True  | Look offline only |
| True  | False  | Look in remote repository only |
| False  | False  | Look in remote repository only and fail |

The path to the remote repo is just a path IMO, and it's least surprising for only the flag to determine whether it's used or not.

mogronalol:
… now always fail if unable to find the stub remotely


marcingrzejszczak:
Shouldn't we fail if we have `workOffline=false` and `repositoryRoot` that is empty?

mogronalol:
In this case I think that's what's tested, "workOffline=false" + "reposRoot=empty", resulting in a failure.  Previously it would have tried to look for the stub locally which I didn't think made sense because you've been explicit about it not doing that by setting `workOffline=false`

marcingrzejszczak:
I'm sorry I'm blind. Disregard my comment

PR #127, opened 2016-10-29
add support for spring-amqp annotated listener endpoints

mduesterhoeft:
I improved the first version of the spring-ampq integration by adding support for multiple listener containers on the application context and also for `@RabbitListener` annotated endpoints.

Let me know what you think.

I am trying to get some feedback from the spring-amqp team via gitter as well.

### DIFF
--- a/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/SpringAmqpStubMessages.java
+++ b/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/SpringAmqpStubMessages.java
@@ -23,6 +23,7 @@ import static org.mockito.Mockito.mockingDetails;
 import static org.mockito.Mockito.verify;
 import static org.springframework.amqp.support.converter.DefaultClassMapper.DEFAULT_CLASSID_FIELD_NAME;
 
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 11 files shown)


spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/SpringAmqpStubMessages.java
```diff
@@ -84,10 +86,14 @@ public SpringAmqpStubMessages(RabbitTemplate rabbitTemplate, MessageListenerAdap
 
 	@Override
 	public void send(Message message, String destination) {
-		if (this.messageListenerAdapter == null) {
-			throw new IllegalStateException("no MessageListenerAdapter wired - cannot send message");
+		List<SimpleMessageListenerContainer> listenerContainers = this.messageListenerAccessor.getListenerContainersForDestination(destination);
+		if (listenerContainers.isEmpty()) {
```
marcingrzejszczak: Now I wonder if we shouldn't throw an exception instead of presenting a warn. If there's no listener then sth's really wrong isn't it? I'll have to check other implementations and see if we shouldn't unify this in all of them. WDYT? For now we can leave it like this and then I'll make another issue to start throwing exceptions

spring-cloud-contract-verifier/src/test/groovy/org/springframework/cloud/contract/verifier/messaging/amqp/MessageListenerAccessorSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+            def listenerContainersForDestination = messageListenerAccessor.getListenerContainersForDestination(exchange)
+        then:
+            listenerContainersForDestination.size() == 1
+            listenerContainersForDestination.get(0) == listenerContainer
+    }
+
+    def givenSimpleMessageListenerContainer() {
+        listenerContainer = new SimpleMessageListenerContainer()
+        listenerContainer.setQueueNames(queueName)
+        binding = BindingBuilder.bind(new Queue(queueName)).to(new DirectExchange(exchange)).with("#")
```
marcingrzejszczak: can you write `return` and types? :P I'm the *statically typed guy that writes returns* kind of a Groovy person :D

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/SpringAmqpStubMessages.java
mduesterhoeft: I agree - the other integrations seem to throw an exception in such a case - I will change this one.

marcingrzejszczak:
Congrats! :)

PR #133, opened 2016-11-06
enable amqp integration by property (#131)

### LINKED ISSUE #131: spring-amqp auto config not applied because spring-integration is pulled when using gradle
Recently I tried the spring-cloud-contract spring-amqp integration in a standalon project using gradle - I realized that spring-cloud-starter-contract-stub-runner pulls in spring-integration. It is optional but gradle does not seem to support that (is that correct?) - in this case the conditionals on `ContractVerifierAmqpAutoConfiguration` do not match because of
```java
@ConditionalOnMissingClass("org.springframework.integration.core.MessageSource")
```

So I ended up having to exclude the spring-integration dependency like so:

```java
testCompile ("org.springframework.cloud:spring-cloud-starter-contract-stub-runner") {
    exclude group: 'org.springframework.integration‘
}
```

Please see the sample project here https://github.com/mduesterhoeft/spring-cloud-contract/tree/feature/amqp-sample/samples/standalone/messaging/amqp-consumer

Currently, I do not have an idea on how to solve this. If we can come up with a solution idea I would be happy to implement it.

mduesterhoeft:
Fixes #131
Changed the conditions to apply the amqp integration auto configuration as follows:

```
@ConditionalOnClass(RabbitTemplate.class)
@ConditionalOnProperty(name="stubrunner.amqp.enabled", havingValue="true")
```

Also changed the property name for the `RabbitMockConnectionFactoryAutoConfiguration` from `verifier.amqp.mockConnection` to `stubrunner.amqp.mockConnection` to use a common prefix.

```
@ConditionalOnProperty(value = "stubrunner.amqp.mockConnection", havingValue = "true", matchIfMissing = true)
```

### DIFF
--- a/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/ContractVerifierAmqpAutoConfiguration.java
+++ b/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/ContractVerifierAmqpAutoConfiguration.java
@@ -20,7 +20,6 @@ import static java.util.Collections.emptyList;
 
 import java.util.List;
 
-import org.mockito.Mockito;
 import org.springframework.amqp.core.Binding;
 import org.springframework.amqp.core.Message;
 import org.springframework.amqp.rabbit.core.RabbitTemplate;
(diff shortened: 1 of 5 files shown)


marcingrzejszczak:
Great work @mduesterhoeft as usual!

PR #151, opened 2016-11-17
Simplification of the dsl

### LINKED ISSUE #149: You shouldn't need to provide a consumer and producer value if you just want to match a pattern
There are many instances where you don't care about the actual value used in the contract, just that it matches a pattern.
I started out with an example sort of like this:
```
request {
	method 'POST'
	url '/thing'
	body(
		id: $(consumer(regex(uuid())), producer('some-really-long-uuid'))
	)
}
```
@marcingrzejszczak informed me that it can be simplified to something like this, by omitting the producer call:

```
request {
	method 'POST'
	url '/thing'
	body(
		id: $(consumer(regex(uuid())))
	)
}
```
If you incorporate the changes from #148, that becomes further simplified to this:

```
request {
	method 'POST'
	url '/thing'
	body(
		id: $(consumer(anyUuid()))
	)
}
```
But if we're not providing a provider value, it would be nice to just imply that it's a consumer value.
That would leave us with something like this:

```
request {
	method 'POST'
	url '/thing'
	body(
		id: $(anyUuid())
	)
}
```
This leaves us with a lot less noise and a much clearer intent. The id field should always be a UUID, no other details required.

marcingrzejszczak:
With this change to DSL we're adding some helper methods

## Consumer / Producer

`c(...)` / `p(...)` - for consumer / producer

Example:

```
Contract contractDsl = Contract.make {
			request {
				method "GET"
				url "test"
			}
			response {
				status 200
				body(
						property1: "a",
						property2: $(
								c('123'),
								p(regex('[0-9]{3}'))
						)
				)
				headers {
					header('Content-Type': 'application/json')
				}

			}
		}
```

so 
- instead of `consumer()` you can write `c()`
- instead of `producer()` you can write `p()`

## any...

example

```
Contract contractDsl = Contract.make {
				request {
					method 'GET'
					urlPath '/get'
					body([
							alpha: $(anyAlphaUnicode()),
							number: $(anyNumber()),
							aBoolean: $(aBoolean()),
							ip: $(anyIpAddress()),
							hostname: $(anyHostname()),
							email: $(anyEmail()),
							url: $(anyUrl()),
							uuid: $(anyUuid())
					])
					headers {
						contentType(applicationJson())
					}
				}
				response {
					status 200
					body([
							alpha: $(anyAlphaUnicode()),
							number: $(anyNumber()),
							aBoolean: $(aBoolean()),
							ip: $(anyIpAddress()),
							hostname: $(anyHostname()),
							email: $(anyEmail()),
							url: $(anyUrl()),
							uuid: $(anyUuid())
					])
					headers {
						contentType(applicationJson())
					}
				}
			}
```

will result in the following test (example for JaxRs)

```
 when:
  def response = webTarget
    .path('/get')
    .request()
    .method('GET', entity('{"alpha":"BRBCPBPEVMLETLINMRAT","number":"1355872605","aBoolean":"true","ip":"252.0d.d.1d","hostname":"http://foo1113543380.com","email":"foo@bar-607214699.com","url":"http://foo-124323798.com","uuid":"49e8ea05-a5b2-4353-b6a4-30981d075577"}', 'application/json'))

  String responseAsString = response.readEntity(String)

 then:
  response.status == 200
  response.getHeaderString('Content-Type') ==~ java.util.regex.Pattern.compile('application/json.*')
 and:
  DocumentContext parsedJson = JsonPath.parse(responseAsString)
  assertThatJson(parsedJson).field("aBoolean").matches("(true|false)")
  assertThatJson(parsedJson).field("alpha").matches("[\\p{L}]*")
  assertThatJson(parsedJson).field("url").matches("((www\\.|(http|https|ftp|news|file)+\\:\\/\\/)[_.a-z0-9-]+\\.[a-z0-9\\/_:@=.+?,##%&~-]*[^.|\\'|\\# |!|\\(|?|,| |>|<|;|\\)])")
  assertThatJson(parsedJson).field("number").matches("-?\\d*(\\.\\d+)?")
  assertThatJson(parsedJson).field("email").matches("[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,4}")
  assertThatJson(parsedJson).field("ip").matches("([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.([01]?\\d\\d?|2[0-4]\\d|25[0-5])")
  assertThatJson(parsedJson).field("uuid").matches("[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}")
  assertThatJson(parsedJson).field("hostname").matches("((http[s]?|ftp):\\/)\\/?([^:\\/\\s]+)(:[0-9]{1,5})?")
```

## HttpHeaders, MediaTypes & HttpMethods

We're introducing most commonly used HTTP Headers, Media Types and HttpMethods. That way you can write

For headers you can use the delegated methods as follows

```
headers {
accept("foo")
contentType("bar")
header(acceptCharset(), textHtml())
}
```

as for HTTP MEthods:

```
request {
method GET()
}
```

instead of 

```
request {
method "GET"
}
```

As you can see here we've removed most of the code in the body of the DSL. Also content type is just one liner with a concrete value and a matcher on consumer / producer sides.

From the developers view this class `org.springframework.cloud.contract.spec.internal.PatternValueDslProperty` is creating all those `all...` methods that basically wrap the `org.springframework.cloud.contract.spec.internal.RegexPatterns` patterns. I've added providing of some default values manually since Xeger (the lib that creates values from the regexps) blows up sometimes if you have unbounded regexps (`[0-9]+`) or sth.

fixes #149

cc @dsyer @dstepanov @Fitzoh

### DIFF
--- a/spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/internal/ClientDslProperty.groovy
+++ b/spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/internal/ClientDslProperty.groovy
@@ -29,4 +29,8 @@ class ClientDslProperty extends DslProperty {
 	ClientDslProperty(Object singleValue) {
 		super(singleValue)
 	}
+
+	ClientDslProperty(Object client, Object server) {
+		super(client, server)
+	}
 }
--- a/spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/internal/Common.groovy
+++ b/spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/internal/Common.groovy
@@ -145,6 +145,17 @@ class Common {
 	}
 
 	/**
+	 * Helper method to provide a better name for the consumer side
+	 */
+	ClientDslProperty c(Object clientValue) {
+		return new ClientDslProperty(clientValue)
+	}
+
+	ServerDslProperty p(Object serverValue) {
+		return new ServerDslProperty(serverValue)
+	}
+
+	/**
 	 * Helper method to provide a better name for the producer side
 	 */
 	ServerDslProperty test(Object serverValue) {
--- a/spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/internal/DslProperty.groovy
+++ b/spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/internal/DslProperty.groovy
@@ -42,4 +42,10 @@ class DslProperty<T> {
 		this.clientValue = singleValue
 		this.serverValue = singleValue
 	}
+
+	boolean isSingleValue() {
+		return this.clientValue == this.serverValue ||
+				(this.clientValue != null && this.serverValue == null ) ||
+				(this.serverValue != null && this.clientValue == null )
+	}
 }
--- a/spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/internal/Headers.groovy
+++ b/spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/internal/Headers.groovy
@@ -30,6 +30,9 @@ import groovy.transform.TypeChecked
 @TypeChecked
 class Headers {
 
+	@Delegate MediaTypes mediaTypes = new MediaTypes()
+	@Delegate HttpHeaders httpHeaders = new HttpHeaders()
+
 	Set<Header> entries = []
 
 	void header(Map<String, Object> singleHeader) {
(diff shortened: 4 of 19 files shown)


dstepanov:
Nice :smile: 
Maybe it would be better to use a reference to the Regexp constants in the test definition instead of in-lining it everywhere. IMHO it would be more readable, but you would require to go into sources to see what pattern is actually used.

marcingrzejszczak:
You can use either of those approaches. You can provide an explicit reference.

PR #161, opened 2016-12-06
Introduction of pluggability

### LINKED ISSUE #154: Add ability to add converters and generators
We need to refactor the core of SC-Contract to allow easier plugability. It would be great to allow easy conversion from STH to DSL and the other way round. Ideally it would be great just to add a dependency with a new converter and it would work OOB.

The same should be done for Test generation and Stub generation.

Related to https://github.com/spring-cloud/spring-cloud-contract/issues/136 and https://github.com/spring-cloud/spring-cloud-contract/issues/129 and https://github.com/spring-cloud/spring-cloud-contract/issues/96

cc @Fitzoh @wesaidso @BenDave

marcingrzejszczak:
with this approach we introduce

- [x]  pluggability of DSL format
- [x]  pluggability of Test generation
- [x]  pluggability of Stub generation
- [x]  multiple contracts in one file
- [x]  pluggability of Stub Runner implementations

TL;DR; thanks to this PR you can set up your contracts in PACT, generate PHP tests and generate stubs in Hoverfly format by passing a JAR with your converter and registering it in `spring.factories`.

# Multiple contracts in one file

It’s possible to define multiple contracts in one file. An example of such a contract can look like this

```

[
        Contract.make {
            name("should post a user")
            request {
                method 'POST'
                url('/users/1')
            }
            response {
                status 200
            }
        },
        Contract.make {
            request {
                method 'POST'
                url('/users/2')
            }
            response {
                status 200
            }
        }
]
```

In this example one contract has the name field and the other doesn’t. This will lead to generation of two tests that will look more or less like this:

```
package org.springframework.cloud.contract.verifier.tests.com.hello;

import com.example.TestBase;
import com.jayway.jsonpath.DocumentContext;
import com.jayway.jsonpath.JsonPath;
import com.jayway.restassured.module.mockmvc.specification.MockMvcRequestSpecification;
import com.jayway.restassured.response.ResponseOptions;
import org.junit.Test;

import static com.jayway.restassured.module.mockmvc.RestAssuredMockMvc.*;
import static com.toomuchcoding.jsonassert.JsonAssertion.assertThatJson;
import static org.assertj.core.api.Assertions.assertThat;

public class V1Test extends TestBase {

	@Test
	public void validate_should_post_a_user() throws Exception {
		// given:
			MockMvcRequestSpecification request = given();

		// when:
			ResponseOptions response = given().spec(request)
					.post("/users/1");

		// then:
			assertThat(response.statusCode()).isEqualTo(200);
	}

	@Test
	public void validate_withList_1() throws Exception {
		// given:
			MockMvcRequestSpecification request = given();

		// when:
			ResponseOptions response = given().spec(request)
					.post("/users/2");

		// then:
			assertThat(response.statusCode()).isEqualTo(200);
	}

}
```

Notice that for the contract that has the name field the generated test method is named `validate_should_post_a_user`. For the one that doesn’t have the name it’s called `validate_withList_1`. It corresponds to the name of the file `WithList.groovy` and the index of the contract in the list.

The generated stubs will look like this

```
should post a user.json
1_WithList.json
```

As you can see the first file got the name parameter from the contract. The second got the name of the contract file `WithList.groovy` prefixed with the index (in this case contract had index `1` in the list of contracts in the file).

# Pluggable architecture

There are cases where you have your contracts defined in other formats for example YAML, RAML or PACT. On the other hand you’d like to profit from the test and stubs generation. It’s really easy to add your own implementation of either of those. Also you can customize the way tests are generated (for example you can generate tests for other languages) and you can do the same for stubs generation (you can generate stubs for other stub http server implementations).

## Custom contract converter

Let’s assume that your contract is written in a YAML file like this:

```
request:
  url: /foo
  method: PUT
  headers:
    foo: bar
  body:
    foo: bar
response:
  status: 200
  headers:
    foo2: bar
  body:
    foo2: bar
```

Thanks to the `ContractConverter` interface  you can register your own implementation of a contract structure converter. Your implementation needs to state the condition on which it should start the conversion. Also you have to define how to perform that conversion in both ways.

Once you create your implementation you have to create a /META-INF/spring.factories file in which you provide the fully qualified name of your implementation.

Example of a spring.factories file

```
# Converters
org.springframework.cloud.contract.spec.ContractConverter=\
org.springframework.cloud.contract.verifier.converter.YamlContractConverter
```

Sample of a YAML implementation

```
@CompileStatic
class YamlContractConverter implements ContractConverter<List<YamlContract>> {

	@Override
	public boolean isAccepted(File file) {
		String name = file.getName()
		return name.endsWith(".yml") || name.endsWith(".yaml")
	}

	@Override
	public Collection<Contract> convertFrom(File file) {
		try {
			YamlContract yamlContract = new Yaml().loadAs(new FileInputStream(file), YamlContract.class)
			return [Contract.make {
				request {
					method(yamlContract?.request?.method)
					url(yamlContract?.request?.url)
					headers {
						yamlContract?.request?.headers?.each { String key, Object value ->
							header(key, value)
						}
					}
					body(yamlContract?.request?.body)
				}
				response {
					status(yamlContract?.response?.status)
					headers {
						yamlContract?.response?.headers?.each { String key, Object value ->
							header(key, value)
						}
					}
					body(yamlContract?.response?.body)
				}
			}]
		}
		catch (FileNotFoundException e) {
			throw new IllegalStateException(e)
		}
	}

	@Override
	public List<YamlContract> convertTo(Collection<Contract> contracts) {
		return contracts.collect { Contract contract ->
			YamlContract yamlContract = new YamlContract()
			yamlContract.request.with {
				method = contract?.request?.method?.clientValue
				url = contract?.request?.url?.clientValue
				headers = (contract?.request?.headers as Headers)?.asStubSideMap()
				body = contract?.request?.body?.clientValue as Map
			}
			yamlContract.response.with {
				status = contract?.response?.status?.clientValue as Integer
				headers = (contract?.response?.headers as Headers)?.asStubSideMap()
				body = contract?.response?.body?.clientValue as Map
			}
			return yamlContract
		}
	}
}
```

## Custom test generator

If you want to generate tests for different languages than Java or you’re not happy with the way we’re building Java tests for you then you can register your own implementation to do that.

Thanks to the `SingleTestGenerator` interface you can register your own implementation that generates a test. Again, it’s enough to provide a proper spring.factories file. Example:

```
org.springframework.cloud.contract.verifier.builder.SingleTestGenerator=/
com.example.MyGenerator
```

## Custom stub generator

If you want to generate stubs for other stub server than WireMock it’s enough to plug in your own implementation of this `SingleFileConverter` interface you can register your own implementation that generate Stubs. Again, it’s enough to provide a proper spring.factories file. Example:

```
# Stub converters
org.springframework.cloud.contract.verifier.converter.SingleFileConverter=\
org.springframework.cloud.contract.verifier.wiremock.DslToWireMockClientConverter
The default implementation is the WireMock stub generation.
```

##  Custom Stub Runner

If you decide to have a custom stub generation you also need a custom way of running stubs with your different stub provider.

Let us assume that you’re using Moco to build your stubs. You wrote a proper stub generator and your stubs got placed in a JAR file.

In order for Stub Runner to know how to run your stubs you have to define a custom HTTP Stub server implementation. It can look like this:

```
	package org.springframework.cloud.contract.stubrunner.provider.moco

import com.github.dreamhead.moco.bootstrap.arg.HttpArgs
import com.github.dreamhead.moco.runner.JsonRunner
import org.springframework.cloud.contract.stubrunner.HttpServerStub
import org.springframework.util.SocketUtils

class MocoHttpServerStub implements HttpServerStub {

	private boolean started
	private JsonRunner runner
	private int port

	@Override
	int port() {
		if (!isRunning()) {
			return -1
		}
		return port
	}

	@Override
	boolean isRunning() {
		return started
	}

	@Override
	HttpServerStub start() {
		return start(SocketUtils.findAvailableTcpPort())
	}

	@Override
	HttpServerStub start(int port) {
		this.port = port
		return this
	}

	@Override
	HttpServerStub stop() {
		if (!isRunning()) {
			return this
		}
		this.runner.stop()
		return this
	}

	@Override
	HttpServerStub registerMappings(Collection<File> stubFiles) {
		List<InputStream> streams = stubFiles.collect { it.newInputStream() }
		this.runner = JsonRunner.newJsonRunnerWithStreams(streams,
				HttpArgs.httpArgs().withPort(this.port).build())
		this.runner.run()
		this.started = true
		return this
	}

	@Override
	boolean isAccepted(File file) {
		return file.name.endsWith(".json")
	}
}
```

and just register it in your `spring.factories` file

```
org.springframework.cloud.contract.stubrunner.HttpServerStub=\
org.springframework.cloud.contract.stubrunner.provider.moco.MocoHttpServerStub
```

that way you’ll be able to run stubs using Moco.

*IMPORTANT*

If you don’t provide any implementation then the default one - WireMock based will be picked. If you provide more than one then the first one on the list will be picked.

fixes #154 

cc @dsyer @Fitzoh  @wesaidso @BenDave @mogronalol


pom.xml
```diff
@@ -136,6 +136,12 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
+			<!-- TODO: Why isn't the version taken from boot bom? -->
+			<dependency>
+				<groupId>org.springframework</groupId>
+				<artifactId>spring-core</artifactId>
+				<version>4.3.4.RELEASE</version>
```
marcingrzejszczak: @dsyer any idea about this? I have to provide the `spring-core` version - otherwise the dependency is not found

dsyer: That can happen if you use the wrong version of Maven. Are you sure you are using the wrapper, and does it have 3.3.9?

dsyer: OK. It's because you explicitly depend on spring-core in spring-cloud-contract-verifier. I suspect it may be adequate to rely on the transitive dependency. I tried simply removing it and the build broke somewhere else, but it seems reasonable to simply remove it in both places to me.

mogronalol:
Brilliant, really liking these changes.

marcingrzejszczak:
@mogronalol - you can easily plug in Hoverfly as a separate dependency with this. I tried to add it in one of the test samples but failed :/ So I used moco

mogronalol:
I think the issue I just realised we are going to have is that it uses Golang regex, so we'd need some sort of Java to Golang regex translator... Sounds nasty and it would be so I'm pretty stumped. Hm.

marcingrzejszczak:
So you can create your own stub generator that converts regex to equality values. And then your server side will understand those. Of course it's not as powerful as with regex but that's better than nothing I guess

mogronalol:
Things like `anyIpAddress()` and different matchers could help aswell.  Because they'd abstract away the regex implementation, and I think it's more expressive anyway.

marcingrzejszczak:
those `anyIpAddress()` are already in 1.0.2.RELEASE. You can also read how to create your own libraries so that you can reuse them in your contracts - https://cloud.spring.io/spring-cloud-contract/1.0.x/spring-cloud-contract.html#_common_jar

mogronalol:
Hm, the regex generated for the client side tests would need to be Java regex, but then when generating the producer stub it would need to be Golang...

marcingrzejszczak:
Ah that sounds complicated :P

Issue #129, opened 2016-11-03
RAML <> SC dsl conversion


wesaidso:
It would be great if I could use our existing RAML files to generate the SC dsl!

marcingrzejszczak:
@wesaidso I've just merge the pluggability thing to master. Now you can easily create the `org.springframework.cloud.contract.spec.ContractConverter` to convert from RAML to DSL and back. Check https://github.com/spring-cloud/spring-cloud-contract/pull/161 for more info

eddumelendez:
Hi @marcingrzejszczak I took this issue and I have some progress converting `raml` files to `Contract` but become harder in the other way. I can send my PR in order to review and continue with this.

marcingrzejszczak:
WOW!! Sure! I'd love to see it!!maybe we'll manage to add it for Dalston :D

marcingrzejszczak:
Following the discussion here https://github.com/spring-cloud/spring-cloud-contract/pull/227 it turns out that schema based solutions are not really suitable as input for Spring Cloud Contract. The contents of the PR will be moved to Spring Cloud Incubator.

PR #193, opened 2017-01-15
Add RAML

eddumelendez:
- [X] Convert Raml to Contract
- [ ]  Convert Contract to Raml
- [ ] Add docs

See #129

### DIFF
--- a/spring-cloud-contract-tools/pom.xml
+++ b/spring-cloud-contract-tools/pom.xml
@@ -22,6 +22,7 @@
 		<module>spring-cloud-contract-spec-pact</module>
 		<module>spring-cloud-contract-maven-plugin</module>
 		<module>spring-cloud-contract-gradle-plugin</module>
+		<module>spring-cloud-contract-spec-raml</module>
 	</modules>
 
 </project>
--- a/spring-cloud-contract-tools/spring-cloud-contract-spec-raml/src/main/groovy/org/springframework/cloud/contract/verifier/spec/raml/RamlContractConverter.groovy
+++ b/spring-cloud-contract-tools/spring-cloud-contract-spec-raml/src/main/groovy/org/springframework/cloud/contract/verifier/spec/raml/RamlContractConverter.groovy
@@ -1,0 +1,180 @@
+/*
+ *  Copyright 2013-2017 the original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.contract.verifier.spec.raml
+
+import groovy.json.JsonSlurper
+import groovy.transform.CompileStatic
+import org.raml.v2.api.RamlModelBuilder
+import org.raml.v2.api.RamlModelResult
+import org.raml.v2.api.model.v10.bodies.Response
+import org.raml.v2.api.model.v10.datamodel.ExampleSpec
+import org.raml.v2.api.model.v10.datamodel.TypeDeclaration
+import org.raml.v2.api.model.v10.methods.Method
+import org.raml.v2.api.model.v10.resources.Resource
+import org.raml.v2.internal.impl.RamlBuilder
+import org.springframework.cloud.contract.spec.Contract
+import org.springframework.cloud.contract.spec.ContractConverter
+import org.springframework.cloud.contract.spec.internal.Headers
+
+/**
+ * Converter of RAML file
+ *
+ * @author Eddú Meléndez
+ * @since 1.1.0
+ */
+@CompileStatic
+class RamlContractConverter implements ContractConverter<RamlBuilder> {
+
+    @Override
+    boolean isAccepted(File file) {
+        RamlModelBuilder ramlContract = new RamlModelBuilder()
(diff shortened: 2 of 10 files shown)


spring-cloud-contract-tools/pom.xml
```diff
@@ -21,7 +21,8 @@
 		<module>spring-cloud-contract-converters</module>
 		<module>spring-cloud-contract-spec-pact</module>
 		<module>spring-cloud-contract-maven-plugin</module>
-		<module>spring-cloud-contract-gradle-plugin</module>
+		<!--<module>spring-cloud-contract-gradle-plugin</module>-->
```
marcingrzejszczak: lol, why? Too slow? :P You can run with `-Pfast` to ignore the tests from that module

spring-cloud-contract-verifier/pom.xml
```diff
@@ -74,6 +74,10 @@
 			<artifactId>groovy-xml</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.raml</groupId>
+			<artifactId>raml-parser-2</artifactId>
+		</dependency>
```
marcingrzejszczak: Mayb let's make this optional?

spring-cloud-contract-verifier/src/main/resources/META-INF/spring.factories
```diff
@@ -5,4 +5,8 @@ org.springframework.cloud.contract.verifier.messaging.integration.ContractVerifi
 org.springframework.cloud.contract.verifier.messaging.amqp.ContractVerifierAmqpAutoConfiguration,\
 org.springframework.cloud.contract.verifier.messaging.amqp.RabbitMockConnectionFactoryAutoConfiguration,\
 org.springframework.cloud.contract.verifier.messaging.camel.ContractVerifierCamelConfiguration,\
-org.springframework.cloud.contract.verifier.messaging.noop.NoOpContractVerifierAutoConfiguration
\ No newline at end of file
+org.springframework.cloud.contract.verifier.messaging.noop.NoOpContractVerifierAutoConfiguration
+
+# Converters
+org.springframework.cloud.contract.spec.ContractConverter=\
+org.springframework.cloud.contract.verifier.converter.RamlContractConverter
```
marcingrzejszczak: What's this?

spring-cloud-contract-tools/spring-cloud-contract-spec-raml/src/main/groovy/org/springframework/cloud/contract/verifier/spec/raml/RamlContractConverter.groovy
```diff
@@ -0,0 +1,180 @@
...
+    @Override
+    Collection<Contract> convertFrom(File file) {
+        RamlModelBuilder ramlBuilder = new RamlModelBuilder()
+        RamlModelResult api = ramlBuilder.buildApi(file)
+        return api.apiV10.resources().collect { Resource resource ->
+            Contract.make {
+                description(resource.description().value())
+                request {
+                    url(resource.relativeUri().value()) {
+                        queryParameters {
```
marcingrzejszczak: Maybe let's add query params only if there are some?

spring-cloud-contract-tools/spring-cloud-contract-spec-raml/src/main/groovy/org/springframework/cloud/contract/verifier/spec/raml/RamlContractConverter.groovy
```diff
@@ -0,0 +1,180 @@
...
+                    }
+
+                    resource.methods().each { Method httpMethod ->
+                        method(httpMethod.method().toUpperCase())
+
+                        httpMethod.body().each { TypeDeclaration requestBody ->
+                            JsonSlurper parser = new JsonSlurper()
+                            if (requestBody.defaultValue()) {
+                                def parsedBody = parser.parseText(requestBody.defaultValue())
+                                if (parsedBody instanceof Map) {
```
marcingrzejszczak: I guess this whole section can be extracted and reused

```
if (parsedBody instanceof Map) {
 +                                    body(parsedBody as Map)
 +                                } else if (parsedBody instanceof List) {
 +                                    body(parsedBody as List)
 +                                } else {
 +                                    body(parsedBody.toString())
 +                                }
```

marcingrzejszczak:
Looks really nice!! So what's the catch with the `Collection<Contract>` -> `RAML` conversion?

marcingrzejszczak:
Ok I think I know what the problem is :P I found this lib that we can use maybe? https://github.com/mulesoft/raml-for-jax-rs, we don't care about the jax-rs part so we can completely exclude it. We can also copy paste their model and parser and reference it

marcingrzejszczak:
Ok I know completely know why you said that it's problematic ;) I've done some of this but I came to the body and now I'm think that we have to choices:

- We'd have to parse it and instead of a value for each node create a type node that defines what that type is
- just set an explicit type like "any"

I actually am in favour of the latter. That's because SC-Contract and contracts as such are not tools to design schemas. Actually what we would present is not really a schema. It's a subset of it. For some M1 I'd suggest we present an MVP without parsing bodies and transforming them into a RAML acceptable structure.

I've pushed my changes here - https://github.com/spring-cloud/spring-cloud-contract/tree/eddumelendez-gh-129 . If you want to continue with this just pick my changes :P

WDYT @eddumelendez ?

marcingrzejszczak:
I have just seen that there's a field in RAML called `example`. I guess that's our only way to pass some request / reply simulation

eddumelendez:
Thanks for your help and suggestions @marcingrzejszczak, I'll take a look and hope to have some deliverable soon.

marcingrzejszczak:
@eddumelendez I sat down a little bit and managed to get some results. You can check out my branch (it's not refactored so it doesn't loot beautiful and has some hardcoded values cause I didn't know what should go in those places) for more info.

marcingrzejszczak:
Closing this in favour of https://github.com/spring-cloud/spring-cloud-contract/pull/227

PR #218, opened 2017-02-07
Fixed unnecessary slash

### LINKED ISSUE #216: WireMockConfiguration incorrectly adding a "/" to files and stubs classpath pattern.
When using a jar to contain wiremock mappings and stub files and specifying a wiremock.server.files location Wiremock fails to locate the mappings.  This is due to org.springframework.cloud.contract.wiremock.WireMockConfiguration.registerFiles() adding a "/" to the end of the path.  
When the WireMockServer constructor is called later WireMockConfiguration.init() it calls the ClassPathFileSource.child() with "__files", which does the following
```
return new ClasspathFileSource(path + "/" + subDirectoryName);
```
This results in "<filesRoot>//__files" which is later not found by the resource loader.  I think the "/" is also not needed in WireMockConfiguration.registerStubs().

I am not sure if this is needed for other FileSource subclasses, but from glancing at the code it does not appear so.

marcingrzejszczak:
without this change the pattern for stubs contains an unnecessary / which corrupts the location search (e.g. /_files is converted to //_files and there is no such path)
with this change we're removing that value

fixes #216

### DIFF
--- a/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockConfiguration.java
+++ b/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockConfiguration.java
@@ -97,9 +97,6 @@ public class WireMockConfiguration implements SmartLifecycle {
 						this.resourceLoader);
 				String pattern = stubs;
 				if (!pattern.contains("*")) {
-					if (!pattern.endsWith("/")) {
-						pattern = pattern + "/";
-					}
 					pattern = pattern + "**/*.json";
 				}
 				for (Resource resource : resolver.getResources(pattern)) {
@@ -115,12 +112,8 @@ public class WireMockConfiguration implements SmartLifecycle {
 			if (StringUtils.hasText(files)) {
 				PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver(
 						this.resourceLoader);
-				String pattern = files;
-				if (!pattern.endsWith("/")) {
-					pattern = pattern + "/";
-				}
 				List<Resource> resources = new ArrayList<>();
-				for (Resource resource : resolver.getResources(pattern)) {
+				for (Resource resource : resolver.getResources(files)) {
 					if (resource.exists()) {
 						resources.add(resource);
 					}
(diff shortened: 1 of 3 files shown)


marcingrzejszczak:
@cortexcompiler can you check if it's working fine for you?

spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockConfiguration.java
```diff
@@ -97,9 +97,6 @@ private void registerStubs() throws IOException {
 						this.resourceLoader);
 				String pattern = stubs;
 				if (!pattern.contains("*")) {
-					if (!pattern.endsWith("/")) {
```
cortexcompiler: Here where the '**/*.json' is being added it is probably Ok to add the "/".

marcingrzejszczak: To tell you the truth I don't really remember why we've added this in the first place. Nor can I break the tests! Do you have some samples that you took that blew up this piece of code? I'm pretty sure you've pasted them somewhere

cortexcompiler: I essentially have my wiremock mappings and stubs in a jar file that is in artifactory.  The stubs are in resources/wiremock/__files and the mappings are in resources/wiremock/mappings.  I am then using: ```@Import(WireMockConfiguration.class)```  along with the following in my application.yml
```
wiremock:
  server:
    files: classpath:wiremock
    stubs: 
    httpsPort: -1
```

cortexcompiler: My issue was primarily with the registerFiles(), since I then could rely on the stubs being picked up from the <filesRoot>/mappings.  If you specify a subs location without the *.json then that code that adds the "/**/*.json" is helpful to find all of the mapping .json files.

cortexcompiler:
I know this is a bit late, but finally got everything working on my Mac.  This change did work for me.
Thanks

PR #220, opened 2017-02-08
Changing the default layout of stubs

### LINKED ISSUE #54: Change default layout of stubs jar files
I think the current layout is too flat, since once a jar is unpacked or inserted on a classpath there is no way to tell the origin of the contracts and mappings. My proposal is `/META-INF/[mappings,contracts]/<groupId>/<artifactId>/<version>/*` instead of `/[mappings,contracts]/*`. Note that WireMock already knows natively how to register stubs from a directory containing `/mappings/**` (that's why we want that at the top of the path).

P.S. I know we don't add stubs to any classpath in the stub runner, but it's a jar file in a Maven repo, so it's easy to do, and could be used to run Wiremock "manually" without the stub runner.

marcingrzejszczak:
without this change the stubs are just placed under the contracts|mappings folders. If you have many of these JARs on the classpath then you might have a problem to say which stubs belong to which server.
with this change we're changing the default location to META-INF/groupid/artifactid/version/contract|mappings . That way you can easily discern stubs

fixes #54


marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-contract/commit/f586cf298d5f6a25b0c4d9237ec7989b4c00123c

PR #227, opened 2017-02-15
Raml as a contract

### LINKED ISSUE #129: RAML <> SC dsl conversion

marcingrzejszczak:
Experimental version of Raml as a contract

TODO:

- [x] add docs
- [x] ensure that RAML files that don't have examples are not used as input

Fixes #129

### DIFF
--- a/docs/src/main/asciidoc/verifier/contract.adoc
+++ b/docs/src/main/asciidoc/verifier/contract.adoc
@@ -822,6 +822,118 @@ include::{standalone_pact_path}/pact-http-client/pom.xml[tags=pact_dependency,in
 include::{standalone_pact_path}/pact-http-client/build.gradle[tags=pact_dependency,indent=0]
 ----
 
+====== RAML converter
+
+WARNING: RAML support is experimental and prone to changes in the future. Also it requires
+ certain conditions on the RAML contract to make the conversion work properly
+
+Spring Cloud Contract comes with an experimental out of the box support for https://raml.org/[RAML] representation of contracts.
+In other words instead of using the Groovy DSL you can use Pact files. In this section
+we will present how to add such a support for your project.
+
+====== RAML constraints
+
+The biggest challenge in making RAML become an input for Spring Cloud Contract is that RAML is
+an extended representation of a schema. It allows you to group a few request examples together
+with some response examples. Spring Cloud Contract doesn't want to bind to a schema thus it
+focuses on presenting a pair of request response examples. That's why not every RAML
+file is possible to be converted into a Spring Cloud Contract example. Of course with time
+and feedback from the users we will try to minimize the number of constraints.
+
+For the RAML contract be considered proper as input for Spring Cloud Contract it has to
+pass the following tests:
+
+- It has to use RAML version 1.0
+- The *baseUri* if present has to contain only the host and no additional path (e.g. *http://localhost:8081* is ok, whereas *http://localhost:8081/api/2.5* is not)
+- Every request that has a body has to have an *example* of a body
+- Every request that has headers has to have *default values* for these headers
+- Every request that has query parameters has to have *default values* for these parameters
+- Every response that has a body has to have an *example* of a body
+- Every response that has headers has to have *default values* for these headers
+
+If you have multiple examples of of bodies only the first one will be taken into consideration.
+The best solution is to have a single pair of request / response in a single RAML file.
+
+====== RAML example
+
+We will be working on the following example of a RAML contract. We've placed this file under
+the `src/test/resources/contracts` folder.
+
+[source,yml,indent=0]
+----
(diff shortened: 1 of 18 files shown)


marcingrzejszczak:
These are the greatest problems when working with RAML as input to SC-Contract. With those in place - is it even possible to use such a RAML file? cc @eddumelendez 

# Challenges

The biggest challenge in making RAML become an input for Spring Cloud Contract is that RAML is
an extended representation of a schema. It allows you to group a few request examples together
with some response examples. Spring Cloud Contract doesn't want to bind to a schema thus it
focuses on presenting a pair of request response examples. That's why not every RAML
file is possible to be converted into a Spring Cloud Contract example. Of course with time
and feedback from the users we will try to minimize the number of constraints.

For the RAML contract be considered proper as input for Spring Cloud Contract it has to
pass the following tests:

- It has to use RAML version 1.0
- The *baseUri* if present has to contain only the host and no additional path (e.g. *http://localhost:8081* is ok, whereas *http://localhost:8081/api/2.5* is not)
- Every request that has a body has to have an *example* of a body
- Every request that has headers has to have *default values* for these headers
- Every request that has query parameters has to have *default values* for these parameters
- Every response that has a body has to have an *example* of a body
- Every response that has headers has to have *default values* for these headers

If you have multiple examples of of bodies only the first one will be taken into consideration.
The best solution is to have a single pair of request / response in a single RAML file.

docs/src/main/asciidoc/verifier/contract.adoc
```diff
@@ -822,6 +822,118 @@ include::{standalone_pact_path}/pact-http-client/pom.xml[tags=pact_dependency,in
 include::{standalone_pact_path}/pact-http-client/build.gradle[tags=pact_dependency,indent=0]
 ----
 
+====== RAML converter
+
+WARNING: RAML support is experimental and prone to changes in the future. Also it requires
+ certain conditions on the RAML contract to make the conversion work properly
+
+Spring Cloud Contract comes with an experimental out of the box support for https://raml.org/[RAML] representation of contracts.
+In other words instead of using the Groovy DSL you can use Pact files. In this section
```
eddumelendez: RAML instead Pact

docs/src/main/asciidoc/verifier/contract.adoc
```diff
@@ -822,6 +822,118 @@ include::{standalone_pact_path}/pact-http-client/pom.xml[tags=pact_dependency,in
...
+
+The biggest challenge in making RAML become an input for Spring Cloud Contract is that RAML is
+an extended representation of a schema. It allows you to group a few request examples together
+with some response examples. Spring Cloud Contract doesn't want to bind to a schema thus it
+focuses on presenting a pair of request response examples. That's why not every RAML
+file is possible to be converted into a Spring Cloud Contract example. Of course with time
+and feedback from the users we will try to minimize the number of constraints.
+
+For the RAML contract be considered proper as input for Spring Cloud Contract it has to
+pass the following tests:
```
eddumelendez: `satisfy the following conditions` sounds much better

docs/src/main/asciidoc/verifier/contract.adoc
```diff
@@ -822,6 +822,118 @@ include::{standalone_pact_path}/pact-http-client/pom.xml[tags=pact_dependency,in
...
+====== RAML converter
+
+WARNING: RAML support is experimental and prone to changes in the future. Also it requires
+ certain conditions on the RAML contract to make the conversion work properly
+
+Spring Cloud Contract comes with an experimental out of the box support for https://raml.org/[RAML] representation of contracts.
+In other words instead of using the Groovy DSL you can use Pact files. In this section
+we will present how to add such a support for your project.
+
+====== RAML constraints
```
eddumelendez: RAML support limitations

eddumelendez:
really nice job @marcingrzejszczak !!! Thanks for continuing with this PR and sorry for the delay.

marcingrzejszczak:
@eddumelendez don't you think that work so many limitations this feature is useless? I have real doubts about merging this. The incompatibility of schemas vs what contact offers is gigantic...

eddumelendez:
@marcingrzejszczak To be honest, yes I think so. Maybe merging this can cause more issues than solutions because users will expect to use their existing RAML files.

marcingrzejszczak:
Ok! I'll move the PR as a separate project to `spring-cloud-incubator` . And I'm not going to merge this to SC-Contract. I'm closing the PR and will close the issue for OpenAPI too.

marcingrzejszczak:
And it's alive :D https://github.com/spring-cloud-incubator/spring-cloud-contract-raml cc @Fitzoh @eddumelendez

PR #235, opened 2017-02-27
use local repository path from maven settings

deki:
currently maven settings are completely ignored

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-build</artifactId>
-		<version>1.2.2.BUILD-SNAPSHOT</version>
+		<version>1.2.2.RELEASE</version>
 		<relativePath/>
 	</parent>
 
--- a/spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/AetherFactories.java
+++ b/spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/AetherFactories.java
@@ -16,7 +16,16 @@
 
 package org.springframework.cloud.contract.stubrunner;
 
+import java.io.File;
+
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -7,7 +7,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-build</artifactId>
-		<version>1.2.2.BUILD-SNAPSHOT</version>
```
marcingrzejszczak: Please go back to SNAPSHOT

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/AetherFactories.java
```diff
@@ -54,7 +63,33 @@ public static RepositorySystemSession newSession(RepositorySystem system, boolea
 	}
 
 	private static String localRepositoryDirectory() {
-		return System.getProperty(MAVEN_LOCAL_REPOSITORY_LOCATION, System.getProperty("user.home") + "/.m2/repository");
+		String localRepoLocationFromSettings = settings().getLocalRepository();
```
marcingrzejszczak: This looks really nice! Could you please add a couple of tests to this?

pom.xml
deki: ok, should be 1.2.3.BUILD-SNAPSHOT, right? 1.2.2 is already released.

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/AetherFactories.java
deki: yes

pom.xml
marcingrzejszczak: Yeah my bad. Please use `1.2.3.BS`

deki:
Opened new PR #236 to 1.0.x branch.

Issue #245, opened 2017-03-17
Extend the DSL with new Client and Server pattern DslProperty creation methods


pfrank13:
When doing some work on my own I came up with a few primitives I think would be generally useful to people at large.  Specifically

1. anyIso8601DateTimeWithTimeZone: adds the Z and +0:00 style endings to a date with optional millisecond precision
2. anyNotEmptyOrWhitespaceString: Anything that is "" or " " will fail
3. enumOf(String...): useful for enumerated values in an API, currencies, languages etc.

I've got no horse in the naming fight, I tend to name verbosely which isn't popular with many people, other suggestions appreciated.

Pull Request forthcoming, I needed a gh issue number to create a proper branch and check messages

marcingrzejszczak:
Hi @pfrank13 ! Great job!

Ensure that any commits are done to a PR for branch `1.0.x` . We're merging then everything from `1.0.x` to master.

As for naming maybe let's go with

```
anyIso8601WithOffset 
anyNonEmptyString
anyOf(...)
```

WDYT? BTW I'm not an expert on naming ;) @Fitzoh maybe you have an opinion on this? :)

Fitzoh:
I like your suggestions @marcingrzejszczak , but I think `anyNonEmptyString` is a little different from `anyNotEmptyOrWhitespaceString`.

Should probably be two options, `anyNonEmptyString` (would allow `' '`) and `anyNonBlankString` (would not allow `' '`).

[isEmpty](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isEmpty-java.lang.CharSequence-)
[isBlank](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isBlank-java.lang.CharSequence-)

marcingrzejszczak:
Oh yeah I've forgotten about the blank method. Obviously you're right. My dementia is getting more severe than I thought.

marcingrzejszczak:
Congrats @pfrank13 on your PR :D

pfrank13:
@marcingrzejszczak Thank you sir :)

PR #247, opened 2017-03-19
gh-245: Created new DslProperty creation methods

pfrank13:
Added:
iso8601WithOffset which handles iso8601 date time with timezone date validation
nonBlankString which handles string that aren't "" and also just whitespace
nonEmptyString which handles "" strings
anyOf(String...) which handles enumerated style attributes for validation

### DIFF
--- a/spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/internal/PatternValueDslProperty.groovy
+++ b/spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/internal/PatternValueDslProperty.groovy
@@ -83,6 +83,23 @@ abstract class PatternValueDslProperty<T extends DslProperty> {
 		return createAndValidateProperty(RegexPatterns.ANY_TIME, "12:2$d:3$d")
 	}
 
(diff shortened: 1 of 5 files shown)


spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/internal/PatternValueDslProperty.groovy
```diff
@@ -83,6 +83,23 @@ abstract class PatternValueDslProperty<T extends DslProperty> {
 		return createAndValidateProperty(RegexPatterns.ANY_TIME, "12:2$d:3$d")
 	}
 
+	T anyIso8601DateTimeWithTimeZone() {
```
marcingrzejszczak: `anyIso8601DateTimeWithTimeZone` should be `anyIso8601WithOffset` ?

pfrank13: Oops my bad, I missed that, it's fixed now

Issue #258, opened 2017-03-30
FAQ update with example configuration for client logging


pfrank13:
From a conversation in Gitter in which I posted a snippet of log4j2.xml to aid in debugging the chatter between the generated tests and the server I was asked to add that to the FAQ and do a PR.  This issue covers that.

PR #259, opened 2017-03-30
gh-258: Added logging config for debugging to FAQ

pfrank13:
Per the conversation in Gitter today, a PR with a FAQ update.  I don't know how to test the generated documentation outside of just looking at the adoc rendering in Intellij, so just a heads up incase I've done something that formats terribly ;)

### DIFF
--- a/docs/src/main/asciidoc/verifier/introduction.adoc
+++ b/docs/src/main/asciidoc/verifier/introduction.adoc
@@ -817,3 +817,27 @@ The rest of the flow looks the same.
 
 Yes! Check out the https://cloud.spring.io/spring-cloud-contract/spring-cloud-contract.html#_different_base_classes_for_contracts[Different base classes for contracts] sections
 of either Gradle or Maven plugins.
+
+==== How can I debug the request/response being sent by the generated tests client?
+
+The generated tests all boil down to RestAssured in some form or fashion which relies on https://hc.apache.org/httpcomponents-client-ga/[Apache HttpClient].  HttpClient has a facility called https://hc.apache.org/httpcomponents-client-ga/logging.html#Wire_Logging[wire logging] which logs the entire request and response to HttpClient.  An example log4j2.xml configurtion could look like this
+[source,xml,indent=0]
+----
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="WARN">
+  <Appenders>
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/verifier/introduction.adoc
```diff
@@ -817,3 +817,27 @@ The rest of the flow looks the same.
 
 Yes! Check out the https://cloud.spring.io/spring-cloud-contract/spring-cloud-contract.html#_different_base_classes_for_contracts[Different base classes for contracts] sections
 of either Gradle or Maven plugins.
+
+==== How can I debug the request/response being sent by the generated tests client?
+
+The generated tests all boil down to RestAssured in some form or fashion which relies on https://hc.apache.org/httpcomponents-client-ga/[Apache HttpClient].  HttpClient has a facility called https://hc.apache.org/httpcomponents-client-ga/logging.html#Wire_Logging[wire logging] which logs the entire request and response to HttpClient.  An example log4j2.xml configurtion could look like this
+[source,xml,indent=0]
+----
+<?xml version="1.0" encoding="UTF-8"?>
```
marcingrzejszczak: Actually I think that it's enough to add in test properties sth like this.

```
logging.level.org.apache.http.wire=DEBUG
```

And then there's no need for any additional logging configuration. WDYT?

Check this out for more info - https://docs.spring.io/spring-boot/docs/current/reference/html/howto-logging.html

pfrank13: Ahhh, I think that is true if Spring Boot is involved.  This may be ignorance on my part, and also the fact that I'm only using the verifier plugin and not doing any of the server side application and generation.

I would say that's a much better approach, would that work if someone is just using the verifier plugin in isolation though?  Guess I can just try and find out =)

pfrank13: I just tried it with my setup which is basically just the plugin and SCC libraries I need and that isn't picked up at all.  Which isn't all that shocking.  I do realize I'm not the intended case for SCC though so I think you're suggestion maps to the intended use case, doing both client and server sides of the equation.  I'm just using the plugin in EXPLICIT mode.

pfrank13:
@marcingrzejszczak, just FYI I made that change you suggested using the Spring Boot property.  Can you please review again?

Issue #270, opened 2017-04-21
Wiremock's SpringHttpServer does not support FaultInjectors


gaqzi:
I have in other projects used the Wiremock [Fault Injectors] to test edge cases for failed requests, for example dump some random data and then close the connection before the request is "finished". 

But with the SpringHttpServer fault injectors are not available. It would be really nice if they were. :)

What I'm trying to do:

```java
        wireMockRule.stubFor(request
                                 .willReturn(aResponse()
                                                 .withFault(Fault.RANDOM_DATA_THEN_CLOSE)
                                 ));
```

The response I get back is `418` with a body of `No fault injector is configured!`.

[Fault Injectors]: http://wiremock.org/docs/simulating-faults/

dsyer:
Do you have a full test case that you can share that would pass if this were supported?

dsyer:
The Wiremock library only supports fault injection with Jetty in native mode (i.e. it doesn't apply for Spring Cloud Contract). We could add support for a `FaultInjectorFactory` to be configured, but the implementation would have to be custom, unless Wiremock can be changed to allow a generic fault injector for all servlet containers. If you have an idea how to implement that, please go ahead and try it.

dsyer:
Actually, it looks like the Jetty version will work in a Spring Boot app. I'd love to have Tomcat and Undertow versions.

dsyer:
Tomcat works with that change (sort of). Probably good enough to move forward.

marcingrzejszczak:
Ok - I'll patch 1.0.x and merge it tomorrow

dsyer:
Merged in d3f6ad9. Please open another ticket for Undertow.

gaqzi:
@dsyer <3

PR #272, opened 2017-04-21
Add a fault injector if server is jetty

dsyer:
It would be good to have the same features for Tomcat and Undertow.

See gh-270.


marcingrzejszczak:
Don't we want this for 1.0.x too?

dsyer:
I guess. Not sure if we should merge before we figure out how to do it with tomcat at least. But go ahead, if you feel like it is too good to miss.

Issue #325, opened 2017-06-13
Maven plugin should reuse Maven's Aether libraries


marcingrzejszczak:
Currently, we transitively take Aether from Stub Runner for Maven plugin.

PR #326, opened 2017-06-13
Shades aether

marcingrzejszczak:
That way we do not interfere with whatever Maven brings on its classpath

without this change deployment might fail due to being unauthorized.
with this change we're shading aether and maven in stub runner. Due to this there is no classpath clash any more

fixes https://github.com/spring-cloud/spring-cloud-contract/issues/325


spring-cloud-contract-starters/spring-cloud-starter-contract-stub-runner/pom.xml
```diff
@@ -33,6 +33,35 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-contract-wiremock</artifactId>
 		</dependency>
+		<!-- for Stub Runner's Aether -->
```
marcingrzejszczak: I'm adding these even though I shouldn't since the dependencies should be shaded... I'll check it out again

spring-cloud-contract-stub-runner/pom.xml
```diff
@@ -172,6 +179,60 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
```
marcingrzejszczak: This should shade the dependencies but for some reason it seems that on the consumer side still original classes are called

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/build.gradle
```diff
@@ -59,6 +59,14 @@ configurations {
 dependencies {
 	compile gradleApi()
 	compile localGroovy()
+	compile "org.eclipse.aether:aether-api:${aetherVersion}"
```
marcingrzejszczak: `api` makes sense since it's not shaded but I also need to pass the rest even though it should get shaded

spring-cloud-contract-starters/spring-cloud-starter-contract-stub-runner/pom.xml
marcingrzejszczak: Now only `api` is on the classpath which is good

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/AetherFactories.java
```diff
@@ -16,27 +16,28 @@
 
 package org.springframework.cloud.contract.stubrunner;
 
+import shaded.org.apache.maven.repository.internal.MavenRepositorySystemUtils;
```
marcingrzejszczak: Here to download a JAR we're using shaded settings

Issue #342, opened 2017-06-26
Stubs generated with RestDocs - JsonMappingException: "No content to map due to end-of-input" when the request has no body


amalianeag:
Hi,
I am generating wiremock stubs with RestDocs. I am trying to test the scenario where no body is specified in the POST request:
```
        mockMvc.perform(post("/{number}/data", 1111).contentType(APPLICATION_JSON_UTF8))
                .andExpect(status().isBadRequest())
                // generating [ WireMock Stub, Contract DSL documentation ] using RestDocs
                .andDo(MockMvcRestDocumentation.document("getData_givenNumberAndNullValues_thenBadRequest", 
                                                         preprocessRequest(prettyPrint()),
                                                         preprocessResponse(prettyPrint()),
                                                         SpringCloudContractRestDocs.dslContract()));
    }
```
I am getting following exception:
```
Caused by: com.fasterxml.jackson.databind.JsonMappingException: No content to map due to end-of-input
 at [Source: ; line: 1, column: 1]
	at com.fasterxml.jackson.databind.JsonMappingException.from(JsonMappingException.java:148)
	at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:3781)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:3721)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2726)
	at com.github.tomakehurst.wiremock.common.Json.read(Json.java:48)
	at com.github.tomakehurst.wiremock.matching.EqualToJsonPattern.<init>(EqualToJsonPattern.java:46)
	at com.github.tomakehurst.wiremock.client.WireMock.equalToJson(WireMock.java:150)
	at org.springframework.cloud.contract.wiremock.restdocs.WireMockSnippet.bodyPattern(WireMockSnippet.java:196)
	at org.springframework.cloud.contract.wiremock.restdocs.WireMockSnippet.requestBuilder(WireMockSnippet.java:167)
	at org.springframework.cloud.contract.wiremock.restdocs.WireMockSnippet.request(WireMockSnippet.java:140)
	at org.springframework.cloud.contract.wiremock.restdocs.WireMockSnippet.document(WireMockSnippet.java:95)
	at org.springframework.restdocs.generate.RestDocumentationGenerator.handle(RestDocumentationGenerator.java:196)
	... 37 more
```
[...]

PR #346, opened 2017-07-02
gh-342: Rest Docs stubs with no body

tkopczynski:
It seems to me that this simple if should resolve the problem described in gh-342


marcingrzejszczak:
Awesome! Thanks a lot!

PR #356, opened 2017-07-13
Adding option to dump mappings to files

### LINKED ISSUE #355: Improve browsing registered stubs
Hi

During my regular work with contracts I often have to resolve some bugs connected with defined contracts (eg 404 caused by missing header ;) ), when I'd like to browse registered stubs I have two options:
a) Add break point in spring-cloud-contract internals to read random generated wire mock ports
b) Add Thread.sleep(100000000) in my business code 
c) Evaluate `curl http://localhost:{wiremockPort}/__admin` (I can't evaluate it without step b) cause thread is blocked)
or
a) read from logs paths with mappings generated in /tmp/..
b) explore generated files

Both of them (in my opinion) are not user friendly, in first approach I have to remember wiremock ports and modify my code (adding thread.sleep()) which is really not cool. In second approach I have to explore my /tmp catalog, after whole day there is a big mess, and I just don't like to explore spring-cloud-contract internals.

My idea to make it more useful is to create temporary file with content similar to response from wiremock `.../__admin` and add this file to `{project}/build/` - just next to generated Spock tests. This file should contain mappings from all set up wiremock instances.

marcingrzejszczak:
without this feature we're missing an option to print all mappings registered for servers
with this feature it will be much easier to debug what was registered in which http server stub. The mappings will be dumped to files with name artifactName_registeredPort

fixes #355

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1339,6 +1339,12 @@ The generated tests all boil down to RestAssured in some form or fashion which r
 logging.level.org.apache.http.wire=DEBUG
 ----
 
+==== How can I see what got registered in the HTTP server stub?
+
+You can use the `mappingsOutputFolder` property on `@AutoConfigureStubRunner` or `StubRunnerRule`
+to dump all mappings per artifact id. Also the port at which the given stub server was
+started will be attached.
+
 ==== Can I reference the request from the response?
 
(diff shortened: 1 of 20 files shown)


pszymczyk:
+1

PR #398, opened 2017-08-29
Fixed NON_EMPTY and NON_BLANK regex patterns

WhyINeedToFillUsername:
The regex patterns NON_BLANK and NON_EMPTY are wrong.
Current NON_EMPTY (".+") does not match a string with a new-line character(s): "\n\n".
Current NON_BLANK (".*(\S+|\R).*|!^\R*$") does not match e.g. "Foo\nBar\n".


marcingrzejszczak:
Please add tests here https://github.com/spring-cloud/spring-cloud-contract/blob/master/spring-cloud-contract-spec/src/test/groovy/org/springframework/cloud/contract/spec/internal/RegexPatternsSpec.groovy

WhyINeedToFillUsername:
Ok, I will take a look at it.

WhyINeedToFillUsername:
updated tests: https://github.com/spring-cloud/spring-cloud-contract/pull/404

marcingrzejszczak:
After updating the PR with tests please ensure that the whole project can be successfully built. For example, I'm more than certain that this test `MockMvcMethodBodyBuilderSpec.should allow an easier way of providing dynamic values for [#methodBuilderName]` will fail. Before you commit your changes please run


```
./mvnw clean install -Pintegration
```

to build the whole project and ensure that you don't break things.

WhyINeedToFillUsername:
I am sorry, I don't think I can get into that right now. I suggested how to fix the patterns and added some true/false test cases. I wanted to make it easier for you by doing a PR, but I can't get into fixing the tests now.

PR #405, opened 2017-08-31
Added composite child support.

tan9:
If specified subDirectoryName exists in more the one FileSource, return another instance of ResourcesFileSource containing all matched FileSources.

### DIFF
--- a/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/file/ResourcesFileSource.java
+++ b/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/file/ResourcesFileSource.java
@@ -34,28 +34,37 @@ import com.github.tomakehurst.wiremock.common.TextFile;
 
 /**
  * @author Dave Syer
- *
+ * @author Pei-Tang Huang
  */
 public class ResourcesFileSource implements FileSource {
 
 	private final FileSource[] sources;
 
-	public ResourcesFileSource(Resource[] resources) {
-		this.sources = new FileSource[resources.length];
+	public ResourcesFileSource(Resource... resources) {
+		this(toSources(resources));
+	}
+
+	private ResourcesFileSource(FileSource... sources) {
+		this.sources = sources;
+	}
+
+	private static FileSource[] toSources(Resource[] resources) {
+		FileSource[] sources = new FileSource[resources.length];
 		for (int i = 0; i < resources.length; i++) {
 			Resource resource = resources[i];
 			if (resource instanceof ClassPathResource) {
 				ClassPathResource classes = (ClassPathResource) resource;
-				this.sources[i] = new ClasspathFileSource(classes.getPath());
+				sources[i] = new ClasspathFileSource(classes.getPath());
 			}
 			else if (resource instanceof FileSystemResource) {
 				FileSystemResource files = (FileSystemResource) resource;
-				this.sources[i] = new SingleRootFileSource(files.getFile());
+				sources[i] = new SingleRootFileSource(files.getFile());
 			}
 			else {
 				throw new IllegalArgumentException("Unsupported resource type for file source: " + resource.getClass());
 			}
 		}
+		return sources;
 	}
 
 	@Override
@@ -94,16 +103,20 @@ public class ResourcesFileSource implements FileSource {
 
 	@Override
 	public FileSource child(String subDirectoryName) {
+		List<FileSource> childSources = new ArrayList<>();
 		for (FileSource resource : this.sources) {
 			try {
 				UrlResource uri = new UrlResource(resource.child(subDirectoryName).getUri());
 				if (uri.createRelative(subDirectoryName).exists()) {
-					return resource.child(subDirectoryName);
+					childSources.add(resource.child(subDirectoryName));
 				}
 			}
 			catch (IOException e) {
 				// Ignore
 			}
+		}
+		if (!childSources.isEmpty()) {
+			return new ResourcesFileSource(childSources.toArray(new FileSource[0]));
 		}
 		return this.sources[0].child(subDirectoryName);
 	}
--- a/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockConfiguration.java
+++ b/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockConfiguration.java
@@ -123,8 +123,10 @@ public class WireMockConfiguration implements SmartLifecycle {
 				}
 			}
 		}
-		ResourcesFileSource fileSource = new ResourcesFileSource(resources.toArray(new Resource[0]));
-		factory.fileSource(fileSource);
+		if (!resources.isEmpty()) {
+			ResourcesFileSource fileSource = new ResourcesFileSource(resources.toArray(new Resource[0]));
+			factory.fileSource(fileSource);
+		}
 	}
 
 	@Override
--- a/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockConfiguration.java
+++ b/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockConfiguration.java
@@ -111,20 +111,20 @@ public class WireMockConfiguration implements SmartLifecycle {
 	}
 
 	private void registerFiles(com.github.tomakehurst.wiremock.core.WireMockConfiguration factory) throws IOException {
+		List<Resource> resources = new ArrayList<>();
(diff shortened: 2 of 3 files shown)


marcingrzejszczak:
Thanks for the pull request! Can you please add some tests to this feature? Also can you explain what you're trying to achieve?

tan9:
Yes, I will add some test and documentation tomorrow.
@marcingrzejszczak  do you think its worth to extract a `CompositeFileSource` super-class?

marcingrzejszczak:
@tan9 I can't really say cause I don't understand what you are doing. You know that you can use the `**` pattern to go through some folders? I mean that if you want to harvest all jsons for any subfolder of foo you can do `**/foo/**/*.json` . There's no need to change any code.

tan9:
@marcingrzejszczak I have a external service providing groups of rest APIs, all I want to do is group stubs more logically, and load them precisely. If my stubs file group like this:

```
stubs ┬- service-1 ┬- function-group-1 ┬- __files
      ...          |                   └- mappings
                   ├- function-group-2 ┬- __files
                   |                   └- mappings
                   ├- function-group-3 ┬- __files
                   |                   └- mappings
                   ...
```

With this patch, I can pick up any sort of stub groups combination while keeping those resources well-organized.

What I needed is actually a `MultiRootFileSource` where WireMock has no built-in implementation.

marcingrzejszczak:
So why not set a pattern **service-1** for the hole service or **service-1/function-group-1/** for a function group inclusion?

marcingrzejszczak:
I can't do it from my mobile. What I meant is to use asterisks to pick either the whole service or a particular function group

marcingrzejszczak:
Ping @tan9 . Why not use the `**` notation to achieve what you want?

tan9:
@marcingrzejszczak Each group of the sutbs/files was captured by the WireMock proxy server, and may be in __conflict__ with other group, so I can not just include all files/stubs all together.

I can use `**` notation to achieve what I want, but the resulting parameters will be more verbose. For example, if I want to pick only function-group-a and function-group-c, the parameter will be somethings like:
```java
(files = {"classpath:/function-group-a/files/**.json", "classpath:/function-group-c/files/**.json"}, stubs = {  "classpath:..", ...})
```

However, if the `child` method returns an aggregated `FileSource` as well, thanks to the WireMock's `__files`, `mappings` sud-dir convention, the above parameter will become:
```java
(files = {"classpath:/function-group-a", "classpath:/function-group-c"})
```

marcingrzejszczak:
So the only thing we gain is that I don't have to write the additional `files/**.json` in the pattern. I'm not convinced. How about you @dsyer ?

dsyer:
I'm not convinced. It's more complex to explain to users too (any patterns are easy to explain and already supported).

tan9:
@marcingrzejszczak , @dsyer I agree with  you. The complexity came from the overloaded `FileSource` abstraction, and the assumption of `mapping` sub-directory of stubs auto-loading.

The best approach seems to educate users to separate `files` and `stubs` from the overloaded FileSource abstraction explicitly. However, it is not possible from Spring Cloud Contract's side. Take a look on README.adoc:
```asciidoc
NOTE: when you configure the `files` root, then it affects the
automatic loading of stubs as well (they come from the root location
in a subdirectory called "mappings")
```

However, when a user specifying multiple file directories (just like me), he/she will suddenly found the above excerpt is incorrect, and only `mappings` in the *first* file source will be included.

marcingrzejszczak:
My suggestion is to close this PR for now, or create a new one to modify the documentation to explain what's not working as it should. If more people ask about this then we could come back to the discussion about adding this feature. WDYT? @tan9 @dsyer ?

tan9:
@marcingrzejszczak After further investigation, I found a bug in the current code base.

How to reproduce on `AutoConfigureWireMockStubsAndMultipleFilesApplicationTests` in `1.1.x` branch:
1. change `files` attribute from `{"classpath:/nonexistent/", "classpath:/root/"}` to `{"classpath:/root/", "classpath:/nonexistent/"}` the test fails. 
    * I made a change in `WireMockConfiguration` to fixed it.
2. change `files` attribute from `{"classpath:/root/, "classpath:/nonexistent/"}` to `{"classpath:/", '"classpath:/root/", "classpath:/nonexistent/"}` the test fails again.  
    * Apply the first commit in this PR, the test case became green again.

tan9:
@marcingrzejszczak Would you please take a look at current commits? To make sure if it meet the requirement.

spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/file/ResourcesFileSource.java
```diff
@@ -34,28 +34,37 @@
...
+	}
+
+	private static FileSource[] toSources(Resource[] resources) {
+		FileSource[] sources = new FileSource[resources.length];
 		for (int i = 0; i < resources.length; i++) {
 			Resource resource = resources[i];
 			if (resource instanceof ClassPathResource) {
 				ClassPathResource classes = (ClassPathResource) resource;
-				this.sources[i] = new ClasspathFileSource(classes.getPath());
+				sources[i] = new ClasspathFileSource(classes.getPath());
```
marcingrzejszczak: Could you please leave `this.` ?

spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/file/ResourcesFileSource.java
```diff
@@ -170,7 +183,7 @@ public boolean exists() {
 	}
 
 	@Override public void deleteFile(String name) {
-
+		throw new UnsupportedOperationException("Resource file sources are read-only");
```
marcingrzejszczak: You want to merge this to `1.1.x` but it's a breaking change. Let's not throw this exception

spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/file/ResourcesFileSource.java
tan9: This method was extracted from the constructor as a static helper method converting resrouce[] to fileSource[], and have no access to the instance field.
Should I rename it to something else, or inline it back to the origin method?

spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/file/ResourcesFileSource.java
tan9: Ok, I will revert this change.

spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/file/ResourcesFileSource.java
marcingrzejszczak: you can call `ResourcesFileSource.this` or pass the `sources[i]` to a `final` variable. We have a checkstyle rule that checks this - that's why I'm asking for this change.

tan9: @marcingrzejszczak , Sorry, I ran checkstyle checks using https://raw.githubusercontent.com/spring-cloud/spring-cloud-build/v1.3.5.RELEASE/spring-cloud-build-tools/src/main/resources/checkstyle.xml , but my CheckStyle-IDEA plugin with CheckStyle 8.2 shows `Checkstyle found no problems in the file(s)` . Am I missing something?

marcingrzejszczak: Interesting... You can see there that there is this `requireThis` entry. I don't know why that doesn't work. Can you try from Maven?

tan9: Result of my `mvn checkstyle:check` run:
image

One more thing, spring-cloud-contract-wiremock's `name` in POM.xml is in a different naming scheme from other modules.

tan9:
The CircleCI [build](https://circleci.com/gh/spring-cloud/spring-cloud-contract/2399?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) failed with following message:
```
Failed to execute goal org.apache.maven.plugins:maven-jar-plugin:3.0.0:jar (default-jar) on project spring-cloud-contract-verifier: Error assembling JAR: Problem creating jar: Execution exception: Java heap space -> [Help 1]
```
Is it related to this PR?

marcingrzejszczak:
No, it's not. I must have misconfigured the memory settings...

PR #429, opened 2017-09-28
Introduced adding of custom WireMock extensions

### LINKED ISSUE #425: Enable custom extensions such as ResponseTemplateTransformer to be registered when using @AutoConfigureStubrunner
I would like to add a custom ResponseTemplateTransformer bean that handles CORS headers both for regular and OPTIONS requests. There are already implementations available in GitHub. Unfortunately, when using @AutoConfigureStubrunner or @EnableStubrunnerServer there is no hook for configuring additional transformers. Internally the ResponseTemplateTransformer is used as the one and only. WireMockSpring.options() cannot be configured to that extent.

So far, our only workaround is to create separate WireMockServer instances on our own that proxy incoming requests to the ones created automatically. We use this for running a stand-alone stub runner server for testing AJAX code, which heavily relies on CORS.

marcingrzejszczak:
without this change we set only a single extension "response-transformer" without an option to provide any other ones
with this change we introduce the "spring.factories" based extension model. It's enough to provide your own implementation of `org.springframework.cloud.contract.verifier.dsl.wiremock.WireMockExtensions` and register it in `spring.factories`. That way you can control all extensions (request / response).

fixes #425

@Buzzardo could you please check out the docs I've added here?
@maverick1601 can you build this locally and check if this suits your needs?

### DIFF
--- a/docs/src/main/asciidoc/verifier_contract.adoc
+++ b/docs/src/main/asciidoc/verifier_contract.adoc
@@ -536,6 +536,33 @@ proper values. Additionally, it registers two helper functions:
 * `escapejsonbody`: Escapes the request body in a format that can be embedded in a JSON.
 * `jsonpath`: For a given parameter, find an object in the request body.
 
+==== Registering your own WireMock extension
+
+WireMock allows to register custom extensions. Spring Cloud Contract by default
+registers the transformer that allows you to reference request from response.
+If you want to provide your own extensions it's enough to register an implementation
+of the `org.springframework.cloud.contract.verifier.dsl.wiremock.WireMockExtensions`
+interface. Since, we're using the `spring.factories` extension approach, just create
+an entry in `META-INF/spring.factories` file similar to this:
+
+[source,groovy,indent=0]
+----
+include::{stubrunner_core_path}/src/test/resources/META-INF/spring.factories[indent=0]
+----
+
+Here you have the example of a custom extension.
+
+.TestWireMockExtensions.groovy
+[source,groovy,indent=0]
+----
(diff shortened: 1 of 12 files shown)


docs/src/main/asciidoc/verifier_contract.adoc
```diff
@@ -536,6 +536,33 @@ proper values. Additionally, it registers two helper functions:
 * `escapejsonbody`: Escapes the request body in a format that can be embedded in a JSON.
 * `jsonpath`: For a given parameter, find an object in the request body.
```
Buzzardo: Hi, Marcin. I rewrote the bit that you added to make it consistent with the rest of the document. You added some good detail there. Thanks. Here's the rewrite:

==== Registering Your Own WireMock Extension

WireMock lets you register custom extensions. By default, Spring Cloud Contract registers 
the transformer, which lets you reference a request from a response. If you want to 
provide your own extensions, you can register an implementation of the
`org.springframework.cloud.contract.verifier.dsl.wiremock.WireMockExtensions` interface. 
Since we use the `spring.factories` extension approach, you can create an entry in 
`META-INF/spring.factories` file similar to the following:

[source,groovy,indent=0]
----
include::{stubrunner_core_path}/src/test/resources/META-INF/spring.factories[indent=0]
----

The following is an example of a custom extension:

.TestWireMockExtensions.groovy
[source,groovy,indent=0]
----
include::{stubrunner_core_path}/src/test/groovy/org/springframework/cloud/contract/verifier/dsl/wiremock/TestWireMockExtensions.groovy[indent=0]
----

IMPORTANT: Remember to override the `applyGlobally()` method and set it to `false` if you
want the transformation to be applied only for a mapping that explicitly requires it.

marcingrzejszczak:
Thanks @Buzzardo for the review!

PR #430, opened 2017-09-30
Update Netflix starters dependencies

eddumelendez:
Spring Cloud Netflix project renamed the starters adding `-netflix-` in
the artifactId. This commit fixes with the new artifacId.


marcingrzejszczak:
I was supposed to do that this week. Thanks!

PR #488, opened 2017-12-08
when contract is ignored, should not generate empty json, empty json is invalid.

chengpohi:
When I work on writing contracts, I have found when I use [Contract.ignored](https://github.com/spring-cloud/spring-cloud-contract/blob/master/spring-cloud-contract-spec/src/main/groovy/org/springframework/cloud/contract/spec/Contract.groovy#L81) to ignore **Contract**, it's still generate the empty stub json file. this is invalid [json format](http://www.ietf.org/rfc/rfc4627.tx), and this also will cause [WireMockStubMapping. buildFrom](https://github.com/spring-cloud/spring-cloud-contract/blob/master/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockStubMapping.java#L9) failed to parse the **stub json** and run the mock server.

```
		if (!request || !response) {
			return ''
		}

		if (groovyDsl.ignored || contract.ignored) {
			return ''
		}
```
https://github.com/spring-cloud/spring-cloud-contract/blob/master/spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/dsl/wiremock/WireMockStubStrategy.groovy#L74 , as the above code, it will return a default empty when contract `ignored`

So this **PR** is trying to fix this by skip to generate the **stub json file** when the generate contract is empty.

### DIFF
--- a/spring-cloud-contract-tools/spring-cloud-contract-converters/src/main/groovy/org/springframework/cloud/contract/verifier/converter/RecursiveFilesConverter.groovy
+++ b/spring-cloud-contract-tools/spring-cloud-contract-converters/src/main/groovy/org/springframework/cloud/contract/verifier/converter/RecursiveFilesConverter.groovy
@@ -95,10 +95,12 @@ class RecursiveFilesConverter {
 						convertedContent.entrySet().eachWithIndex { Map.Entry<Contract, String> content, int index ->
 							Contract dsl = content.key
 							String converted = content.value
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
This is great! Can you apply this to branch `1.2.x` so that I will then merge it to master?

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-contract/commit/bff685eb46062609b50fa333efe9bf7e0739e8e0

Issue #473, opened 2017-11-28
Slice Testing with just the Autoconfigurations related to stubrunner loaded in


bijukunjummen:
As a consumer of a service with stubs published, it would be great to load up only the spring configuration and components related to stub runner without needing to load up the entire application context. For a complex app with its own set of auto-configurations and with external database dependencies, loading up an entire context is difficult.

Just to provide a little more context, a consumer test looks like this:

```java
@RunWith(SpringRunner::class)
@SpringBootTest
@AutoConfigureStubRunner(ids = arrayOf("samples.contract:sample-producer:+:stubs"), workOffline = true)
class PassThroughContractTest {
```

The `@SpringBootTest` annotation causes the entire boot application context along with any auto-configurations to be loaded up which is costly for na app with likely external dependencies.

Junit rule works well as a substitute. It would be great to have a more Spring Boot native solution also.

Here is the project where I was trying it out - https://github.com/bijukunjummen/s-c-contract-sample

bijukunjummen:
I will try and send you a PR @marcingrzejszczak

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-contract/pull/496

PR #496, opened 2017-12-15
Support loading limited context for stub runner tests

bijukunjummen:
#473 - @marcingrzejszczak , this is still Work in progress, but I wanted to send you a PR to see if this directionally looks okay. Can you please review when you get a chance.


marcingrzejszczak:
This looks very good! I'd only add `@since 2.0.0` in the javadocs and add some documentation. Anything you would add @dsyer ?

dsyer:
We prefer test classes to be called `*Tests`. That should be easy to change.

dsyer:
I don't really understand the problem this solves, so some documentation will be helpful as well. Otherwise there will be a lot of confusion about why this is needed in addition to `@AutoConfigureStubRunner` (is it, actually, needed?).

marcingrzejszczak:
My understanding is that you don't need the full Spring Boot context to use StubRunner thanks to this, am I correct @bijukunjummen ?

dsyer:
Is there a way to just use `@AutoconfigureStubRunner` in a normal (existing) test slice?

bijukunjummen:
@marcingrzejszczak , @dsyer , I have now removed the slice test for stub runner and instead modified like you suggested `@AutoConfigureStubRunner` to load only the relevant auto-configurations, this way it works in combination with any other test slice.

marcingrzejszczak:
Thanks a lot @bijukunjummen !

PR #510, opened 2018-01-07
replaced ".class" with ".getClass()"

### LINKED ISSUE #509: NullPointerException when trying to create producer tests from Pact
I'm trying to auto-create producer tests from a Pact contract using `spring-cloud-contract-spec-pact` and the verifier in version 1.2.2.BUILD-SNAPSHOT (but I reproduced the error in 1.2.1 as well).

Each time the gradle task `generateContractTests` fails with the following stacktrace:

```
...
Caused by: java.lang.NullPointerException: Cannot get property 'name' on null object
        at org.springframework.cloud.contract.verifier.builder.MethodBodyBuilder.checkType(MethodBodyBuilder.groovy:464)
        at org.springframework.cloud.contract.verifier.builder.MethodBodyBuilder.methodForTypeCheck(MethodBodyBuilder.groovy:422)
        at org.springframework.cloud.contract.verifier.builder.MethodBodyBuilder$_addJsonResponseBodyCheck_closure3.doCall(MethodBodyBuilder.groovy:365)
        at org.springframework.cloud.contract.verifier.builder.MethodBodyBuilder.addJsonResponseBodyCheck(MethodBodyBuilder.groovy:359)
        at org.springframework.cloud.contract.verifier.builder.MethodBodyBuilder.validateResponseBodyBlock(MethodBodyBuilder.groovy:320)
...
```

Here's the pact contract I was using:

```json
{
  "consumer": {
    "name": "ui"
  },
  "provider": {
    "name": "userservice"
  },
  "interactions": [
    {
      "description": "a request to POST a person",
      "providerState": "provider accepts a new person",
      "request": {
        "method": "POST",
        "path": "/user-service/users",
        "headers": {
        },
        "body": {
          "firstName": "Arthur",
          "lastName": "Dent"
        }
      },
      "response": {
        "status": 201,
        "headers": {
        },
        "body": {
          "id": 42
        },
        "matchingRules": {
          "$.body": {
            "match": "type"
          }
        }
      }
    }
  ],
  "metadata": {
    "pactSpecification": {
      "version": "2.0.0"
    }
  }
}
```

**Further analysis:**
I traced the error back to `MethodBodyBuilder.classToCheck()` where the statement

```java
switch (elementFromBody.class) {
```

should probably be replaced with 
```java
switch (elementFromBody.getClass()) {
```

If `elementFromBody` is a `Map` (as in my case), `elementFromBody.class` in Groovy apparently resolves to the element with the key `class` within the map (which is null) instead of the class of the Map itself.

If the statement above is replaced, the test is generated successfully and evens runs successfully.

thombergs:
so that the actual class of the object is returned instead of
potentially an element with key "class" if object is a Map

fixes #509


thombergs:
I haven't added a test since I do not know how to create a contract with the DSL that goes into the last `else` branch (`MethodBodyBuilder.groovy` line 365; see snippet below). Originally I produced the error with a contract converted from a pact file.

```groovy
bodyMatchers.jsonPathMatchers().each {
				if (MatchingType.regexRelated(it.matchingType()) || it.matchingType() == MatchingType.EQUALITY) {
					methodForEqualityCheck(it, bb, copiedBody)
				} else if (it.matchingType() == MatchingType.COMMAND) {
					methodForCommandExecution(it, bb, copiedBody)
				} else {
					methodForTypeCheck(it, bb, copiedBody)
				}
			}
```

thombergs:
I actually believe that this error cannot be produced with creating a contract via DSL. 

I suspect there is a bug in the pact-to-contract mapping that makes the verifier try a type matching against a whole json object like
```json
{ id: 42 }
```
instead of against the single field (`id` in this case). Thus, the type match is done against a `LinkedHashMap` which produces the error.

I will investigate in the pact mapping mechanism and file another issue.

In the meantime, I added a unit test that fixes the bug on a method level instead of on a contract level.

marcingrzejszczak:
That's fantastic! One question though... Could you please apply this PR to the `1.2.x` branch? I would then merge `1.2.x` to `master`.

thombergs:
done

marcingrzejszczak:
Sth looks wrong. Can you close this PR. Create a new branch from `1.2.x` in your fork and create a PR from that new branch? I guess you haven't changed 358 files ;)

thombergs:
no, I did not change 358 files :)
I changed the base for the PR via GitHub UI. But that obviously included other commits than only mine. 

Will open a new PR...

PR #513, opened 2018-01-10
replaced ".class" with ".getClass()"

### LINKED ISSUE #509: NullPointerException when trying to create producer tests from Pact

thombergs:
... so that the actual class of the object is returned instead of
potentially an element with key "class" if object is a Map.

fixed #509

### DIFF
--- a/spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/MethodBodyBuilder.groovy
+++ b/spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/MethodBodyBuilder.groovy
@@ -468,7 +468,7 @@ abstract class MethodBodyBuilder {
 
 	// we want to make the type more generic (e.g. not ArrayList but List)
 	protected Class classToCheck(Object elementFromBody) {
-		switch (elementFromBody.class) {
+		switch (elementFromBody.getClass()) {
 			case List:
 				return List
 			case Set:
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
Thanks a lot!!

PR #519, opened 2018-01-12
replace invalid characters in generated method names with underscore

### LINKED ISSUE #518: contract file names with special chars lead to compile errors in generated tests
If a contract file name contains a character that is invalid in a Java method name (like '-') the generated test will have a compile error since the name of the contract file is taken as a method name without converting invalid characters.

thombergs:
fixes #518


marcingrzejszczak:
@thombergs you're on 🔥 !!

PR #569, opened 2018-03-07
Support for Pact v3

### LINKED ISSUE #583: Usage of MatchingRules for response header
I am using a PACT file of type JSON to test my API. I want to assert the value of a response header which is returned by my API. The value of the response header is an UUID automatically generated on every call. Hence, I need to use matchingRules with regex. 

My Pact File:

```
{
  "metadata": {
    "pact-specification": {
      "version": "3.0.0"
    },
    "pact-jvm": {
      "version": "3.5.4"
    }
  },
  "interactions": [
    {
      "description": "some description",
      "request": {
        "method": "POST",
        "path": "/v1/accounts",
        "body": {
          "nickname": "Nickname"
        },
        "headers": {
          "Accept": "application/json",
        }
      },
      "response": {
        "status": 201,
        "headers": {
          "Content-Type": "application/vnd.fraud.v1+json;charset=UTF-8"
        },
        "body": {
          "nickname": "Nickname"
        },
        "matchingRules": {
          "$.header.something": {
            "match": "regex",
            "regex": "[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])T(2[0-3]|[01][0-9]):[0-5][0-9]:[0-5][0-9].*"
          },
          "$.body.createdDate": {
            "match": "regex",
            "regex": "[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])T(2[0-3]|[01][0-9]):[0-5][0-9]:[0-5][0-9].*"
          }
        }
      }
    }
  ]
}
```

My maven config:

```
<plugin>
                <artifactId>spring-cloud-contract-maven-plugin</artifactId>
                <version>1.2.1.RELEASE</version>
                <extensions>true</extensions>
                <configuration>
                    <packageWithBaseClasses>com.example.contract</packageWithBaseClasses>
                </configuration>
                <dependencies>
                    <dependency>
                        <artifactId>spring-cloud-contract-spec-pact</artifactId>
                        <version>1.2.2.RELEASE</version>
                    </dependency>
                    <dependency>
                        <groupId>au.com.dius</groupId>
                        <artifactId>pact-jvm-model</artifactId>
                        <version>2.4.18</version>
                    </dependency>
                </dependencies>
            </plugin>

```

The generated Java class file does not assert the header with regex.

### LINKED ISSUE #595: JSON value in the Request Header is not transformed correctly
The current implementation of spring-clould-contract does not allow the usage of JSON values in the request header. E.g 

PACT File:
```
"headers": {
          "Accept": "application/json",
          "someHeader": "{\"key\":\"value\"}"
}
```
Java Class:

```
 MockMvcRequestSpecification request = given()
					.header("Accept", "application/json")
					.header("someHeader", "[key:value]");
```
 

The { becomes [ and the " are vanished.

Expected:

```
 MockMvcRequestSpecification request = given()
					.header("Accept", "application/json")
					.header("someHeader", "{\"key\":\"value\"}");
```

TYsewyn:
PR to track the Pact v3 support.

❗️This is currently a work in progress❗️

**Tasks**

- [x] Upgraded `pact-jvm-model` from `2.4.18` to `3.5.13`
- [x] Uses the Pact DSL (`pact-jvm-consumer-java8_2.12`)
- [x] [Query strings stored as map](https://github.com/pact-foundation/pact-specification/tree/version-3#query-strings-are-stored-as-map-instead-of-strings)
- [x] [Multiple providers states in description](https://github.com/pact-foundation/pact-specification/tree/version-3#allow-multiple-provider-states-with-parameters)
- [x] [Multiple matchers for the same path](https://github.com/pact-foundation/pact-specification/tree/version-3#allow-arrays-of-matchers-to-be-defined-against-a-matcher-path)
- [x] [Messaging](https://github.com/pact-foundation/pact-specification/tree/version-3#introduces-messages-for-services-that-communicate-via-event-streams-and-message-queues)
- [x] [Path, Query & Header matchers](https://github.com/pact-foundation/pact-specification/tree/version-3#drop-the-jsonpath-notation-for-matchers)
  - [Note on Content-Type header matching](https://github.com/pact-foundation/pact-specification/tree/version-3#content-type-header-matching-should-include-parameters-in-the-matching)
- [x] [New matchers](https://github.com/pact-foundation/pact-specification/tree/version-3#add-an-equality-matcher)
- [x] [Example generators](https://github.com/pact-foundation/pact-specification/tree/version-3#introduce-example-generators)
- [x] Adapt documentation

**Notes on implementation:**
- Only supports the AND rule logic to combine multiple matchers for the same json path
- Only the body and header matchers are currently supported, path and query matchers will be added in a future release.

**Introduces breaking changes:**
- Returns a collection of Pact contracts instead of one (see https://github.com/spring-cloud/spring-cloud-contract/pull/569/commits/d942bfd9d127489c447ebae3221e063099420492#diff-3c9a6ccf897ca28d932eb2bef60dc2e9R27)
- ~The `testMatchers` and `stubMatchers` have been changed to be able to add path, query and header matchers~ This has been reverted in favor of a non-invasive change using deprecation

Closes #583 
Fixes #595


marcingrzejszczak:
Apparently in Pact 2.0 support, I haven't noticed that in the Pact matchers section you could do `$.body` and pass a regular expression, or `$.headers` and match a header. I haven't noticed the `headers` part :/

marcingrzejszczak:
Related - https://github.com/spring-cloud/spring-cloud-contract/issues/583

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/pom.xml
```diff
@@ -35,8 +35,7 @@
 		</dependency>
 		<dependency>
 			<groupId>au.com.dius</groupId>
-			<artifactId>pact-jvm-model</artifactId>
-			<optional>true</optional>
+			<artifactId>pact-jvm-consumer-java8_2.12</artifactId>
```
marcingrzejszczak: Why not optional? Do we want to force that particular version?

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/BodyConverter.groovy
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	static DslPart toPactBody(DslProperty dslProperty, Closure dslPropertyValueExtractor) {
+		return traverse(dslProperty, null, dslPropertyValueExtractor)
+	}
+
+	private static DslPart traverse(Object value, DslPart parent, Closure dslPropertyValueExtractor) {
+		boolean isRoot = parent == null
+		Object v = value
+
```
marcingrzejszczak: It's just a matter of preference, but could you get rid of all of those separations between blocks? If not, then I'll do it later

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/BodyConverter.groovy
```diff
@@ -0,0 +1,130 @@
...
+			if (!isRoot) {
+				p = p.array()
+			}
+			processCollection(v as Collection, p as PactDslJsonArray, dslPropertyValueExtractor)
+			if (!isRoot) {
+				p = p.closeArray()
+			}
+		}
+
+		p
```
marcingrzejszczak: Can you add `return` ? I prefer to always explicitly define `return`, otherwise you can accidentally return sth you didn't want to return, and also it's more comprehensible for others. Also if you refactor from a method that returns sth to `void`, your code will still compile even though it shouldn't :)

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/BodyConverter.groovy
```diff
@@ -0,0 +1,130 @@
...
+				jsonArray.string(v)
+			} else if (v instanceof Number) {
+				jsonArray.number(v)
+			} else {
+				traverse(it, jsonArray, dslPropertyValueExtractor)
+			}
+		})
+	}
+
+	private static void processMap(Map<String, Object> values, PactDslJsonBody jsonObject, Closure dslPropertyValueExtractor) {
```
marcingrzejszczak: Isn't iterating JSON a fantastic experience? :D

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MatchingRulesConverter.groovy
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * @author Tim Ysewyn
+ * @Since 2.0.0
+ */
+@CompileStatic
+@PackageScope
+class MatchingRulesConverter {
+
+	static Category matchingRulesForBody(BodyMatchers bodyMatchers) {
+		matchingRulesFor("body", bodyMatchers)
```
marcingrzejszczak: `return` please

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MatchingRulesConverter.groovy
```diff
@@ -0,0 +1,68 @@
...
+						category.setRule(key, new MinMaxTypeMatcher(it.minTypeOccurrence(), it.maxTypeOccurrence()))
+					} else if (it.minTypeOccurrence()) {
+						category.setRule(key, new MinTypeMatcher(it.minTypeOccurrence()))
+					} else if (it.maxTypeOccurrence()) {
+						category.setRule(key, new MaxTypeMatcher(it.maxTypeOccurrence()))
+					} else {
+						category.setRule(key, TypeMatcher.INSTANCE)
+					}
+					break
+				case MatchingType.DATE:
```
marcingrzejszczak: There are no longer such rules in Pact? I added them specifically cause they were in pact.

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MessagePactCreator.groovy
```diff
@@ -0,0 +1,98 @@
...
+
+				messagePactBuilder = messagePactBuilder.withContent(pactResponseBody)
+			}
+
+			if (message.headers) {
+				messagePactBuilder = messagePactBuilder.withMetadata(getMetadata(message.headers))
+			}
+		}
+
+		messagePactBuilder.toPact()
```
marcingrzejszczak: `return`

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MessagePactCreator.groovy
```diff
@@ -0,0 +1,98 @@
...
+			OutputMessage message = contract.outputMessage
+			"message sent to " + clientValueExtractor.call(message.sentTo)
+		} else {
+			"assert that " + contract.input.assertThat.executionCommand
+		}
+	}
+
+	private Map<String, String> getMetadata(Headers headers) {
+		headers.entries.collectEntries({ Header header ->
+			["$header.name": extractValue(header)]
```
marcingrzejszczak: Your Groovy FOO is strong! :)

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MessagingSCContractCreator.groovy
```diff
@@ -0,0 +1,136 @@
...
+import org.springframework.cloud.contract.verifier.util.JsonToJsonPathsConverter
+
+/**
+ * Creator of {@link Contract} instances
+ *
+ * @author Tim Ysewyn
+ * @since 2.0.0
+ */
+@CompileStatic
+@PackageScope
```
marcingrzejszczak: BTW Thanks for making everything compile static with package scope!

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MessagingSCContractCreator.groovy
```diff
@@ -0,0 +1,136 @@
...
+					}
+				}
+
+				outputMessage {
+					if (message.contents.present) {
+						body(parseBody(message.contents))
+
+
+						Category bodyRules = message.matchingRules.rulesForCategory('body')
+						if (bodyRules && !bodyRules.matchingRules.isEmpty()) {
```
marcingrzejszczak: matching rules can't be null ? I smell NPE ;)

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MessagingSCContractCreator.groovy
```diff
@@ -0,0 +1,136 @@
...
+							testMatchers {
+								bodyRules.matchingRules.each { String key, MatchingRuleGroup ruleGroup ->
+									if (FULL_BODY.equals(key)) {
+										JsonPaths jsonPaths = JsonToJsonPathsConverter.transformToJsonPathWithStubsSideValuesAndNoArraySizeCheck(message.contents.value)
```
marcingrzejszczak: I apologise for the contents of the `JsonToJsonPathsConverter` class 😭

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MessagingSCContractCreator.groovy
```diff
@@ -0,0 +1,136 @@
...
+										jsonPaths.each {
+											jsonPath(it.keyBeforeChecking(), byType())
+										}
+									} else {
+										ruleGroup.rules.each { MatchingRule rule ->
+											if (rule instanceof RegexMatcher) {
```
marcingrzejszczak: Theoretically, you could rewrite this to call `jsonPath` only once and just pass different value basing on the instance of `rule`.

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/PactContractConverter.groovy
```diff
@@ -52,284 +37,32 @@ class PactContractConverter implements ContractConverter<Pact> {
...
+		})
 
-	protected QueryParameters queryParams(Contract dsl) {
-		if (dsl.request.urlPath) {
-			return dsl.request.urlPath.queryParameters
-		} else if (dsl.request.url) {
-			return dsl.request.url.queryParameters
-		}
-		throw new IllegalStateException("No url provided")
+		pactContracts
```
marcingrzejszczak: missing `return` (I wasn't adding this comment everywhere since it doesn't make much sense to repeat myself)

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/RequestResponsePactCreator.groovy
```diff
@@ -0,0 +1,166 @@
...
+ * Creator of {@link RequestResponsePact} instances
+ *
+ * @author Tim Ysewyn
+ * @since 2.0.0
+ */
+@CompileStatic
+@PackageScope
+class RequestResponsePactCreator {
+
+	static final Closure requestDslPropertyValueExtractor = { DslProperty property -> property.serverValue }
```
marcingrzejszczak: ha! you have to add a modifier e.g. `protected` if you want this to be a constant field. Otherwise if you check out the MOP, you'll see that a `private static final` field was created with a `public static getter` method :)

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/test/groovy/org/springframework/cloud/contract/verifier/spec/pact/PactContractConverterSpec.groovy
```diff
@@ -185,12 +193,18 @@ class PactContractConverterSpec extends Specification {
           "something": "foo"
         },
         "matchingRules": {
-          "$.body.id": {
-            "match": "regex",
-            "regex": "[0-9]{3}"
-          },
-          "$.body.something": {
-            "match": "equality"
+          "body": {
```
marcingrzejszczak: maybe we should check for the headers matchers too?

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/test/groovy/org/springframework/cloud/contract/verifier/spec/pact/PactContractConverterSpec.groovy
```diff
@@ -303,22 +330,245 @@ class PactContractConverterSpec extends Specification {
...
+							})
+							jsonPath('$.data.array1', byType() {
+								minOccurrence(0)
+							})
+							jsonPath('$.data.array2', byType() {
+								minOccurrence(1)
+							})
+							jsonPath('$.id', byType())
+
+//							TODO: CURRENTLY NOT SUPPORTED
```
marcingrzejszczak: Why isn't it supported? That should work I guess OOB ?

TYsewyn: No, it wasn't. They added new matchers to v3, check https://github.com/pact-foundation/pact-specification/tree/version-3#add-an-equality-matcher until the generators

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/pom.xml
TYsewyn: My bad, that wasn't supposed to happen. Then again, I won't know if it will work with future versions.
(except when somebody files an issue)

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/BodyConverter.groovy
TYsewyn: Consider it done

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/BodyConverter.groovy
TYsewyn: Consider it done

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/BodyConverter.groovy
TYsewyn: Oh yes, really fantastic! I'm glad I can mention this on my resume now :D

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MessagingSCContractCreator.groovy
TYsewyn: No, the matchingRules parameter will be set by default on the `Message` object.
But see this [comment](https://github.com/spring-cloud/spring-cloud-contract/pull/569/files/bb44f55cf028564e82f0ab6222a67d0627d56087#r175776041)

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MessagingSCContractCreator.groovy
TYsewyn: Good idea! But then you would need to explain me how I could pass the current closure context to another method.
Closures, they're possibly great when you understand it completely. :D

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MessagingSCContractCreator.groovy
TYsewyn: No problem, just look at it as room for improvement.
I do that on a daily basis ;)

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/RequestResponsePactCreator.groovy
TYsewyn: Good catch! Both should be private indeed!

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/test/groovy/org/springframework/cloud/contract/verifier/spec/pact/PactContractConverterSpec.groovy
TYsewyn: Yes, that's for some time later. They also introduced path & query matchers.
See https://github.com/pact-foundation/pact-specification/tree/version-3#drop-the-jsonpath-notation-for-matchers

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MessagingSCContractCreator.groovy
marcingrzejszczak: Ah ok, so just ignore it.

spring-cloud-contract-tools/spring-cloud-contract-spec-pact/src/main/groovy/org/springframework/cloud/contract/verifier/spec/pact/MatchingRulesConverter.groovy
TYsewyn: Oh they are, for some reason I looked over it.
And I will need to add 4 more for v3: INCLUDE, NUMBER, DECIMAL and NULL

PR #571, opened 2018-03-11
Added anyDouble DslProperty

axelhodler:
Currently were copy and pasting a lot of regexes for doubles throughout our services and the corresponding contracts. Maybe having this could be helpful for others too. (Was super glad someone already added `anyIso8601WithOffset`)

A few sidenotes:

Since the change is a lot more invasive than I initially thought. Is there a suggested way to extend the functionality without changing the library? E.g. we use the type coordinate (lat, lng) quite a lot. But thats probably not generic enough to be added to spring-cloud-contract.

Would there be an interest in `anyHttpsUrl()` (current one only creates `http:`) and `anyPositiveNumber()` (looks weird to the stubrunner consumers if they get like `"pages": -472498` in a response.

### DIFF
--- a/spring-cloud-contract-verifier/src/test/groovy/org/springframework/cloud/contract/verifier/builder/JaxRsClientMethodBuilderSpec.groovy
+++ b/spring-cloud-contract-verifier/src/test/groovy/org/springframework/cloud/contract/verifier/builder/JaxRsClientMethodBuilderSpec.groovy
@@ -1008,6 +1008,7 @@ class JaxRsClientMethodBuilderSpec extends Specification implements WireMockStub
 					body([
 							alpha: $(anyAlphaUnicode()),
 							number: $(anyNumber()),
+							aDouble: $(anyDouble()),
 							aBoolean: $(aBoolean()),
 							ip: $(anyIpAddress()),
 							hostname: $(anyHostname()),
@@ -1031,6 +1032,7 @@ class JaxRsClientMethodBuilderSpec extends Specification implements WireMockStub
 					body([
 							alpha: $(anyAlphaUnicode()),
 							number: $(anyNumber()),
+							aDouble: $(anyDouble()),
 							aBoolean: $(aBoolean()),
 							ip: $(anyIpAddress()),
 							hostname: $(anyHostname()),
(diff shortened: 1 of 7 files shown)


spring-cloud-contract-verifier/src/test/groovy/org/springframework/cloud/contract/verifier/builder/JaxRsClientMethodBuilderSpec.groovy
```diff
@@ -1008,6 +1008,7 @@ class JaxRsClientMethodBuilderSpec extends Specification implements WireMockStub
 					body([
 							alpha: $(anyAlphaUnicode()),
 							number: $(anyNumber()),
+							double: $(anyDouble()),
```
marcingrzejszczak: shouldn't this be in quotes `"double"` since it's a keyword?

spring-cloud-contract-verifier/src/test/groovy/org/springframework/cloud/contract/verifier/builder/MockMvcMethodBodyBuilderSpec.groovy
```diff
@@ -1986,6 +1986,7 @@ World.'''"""
 						duck: $(regex("[0-9]")),
 						alpha: $(anyAlphaUnicode()),
 				        number: $(anyNumber()),
+						double: $(anyDouble()),
```
marcingrzejszczak: Again shouldn't it be in quotes?

marcingrzejszczak:
Also, I think we should add this into `2.0.x` since I would prefer not to add any more features to `1.2.x` branch

spring-cloud-contract-verifier/src/test/groovy/org/springframework/cloud/contract/verifier/builder/JaxRsClientMethodBuilderSpec.groovy
axelhodler: worked without the quotes, but i have changed it to `aDouble` (as its already using `aBoolean`) to avoid confusion

axelhodler:
Thanks!

I will fix the conflicts and change the target branch to master

axelhodler:
rebased latest master and fixed build

PR #574, opened 2018-03-12
Fix disabling of publish task when env property is set

alek-sys:
Because of Gradle's project lifecycle, disabling tasks in the script doesn't
work as expected, ./gradlew publish will still try to execute
:publishMavenJavaPublicationToMavenRepository and fail even if PUBLISH_ARTIFACTS
is set to false. To avoid that and make sure all publish tasks are disabled
script need to listen for tasks graph event to be ready and then disable the tasks.


marcingrzejszczak:
Oh man, great catch! Thanks so much. I'll backport it to 1.2.x

marcingrzejszczak:
Congrats on your first time contribution! :)

Issue #580, opened 2018-03-14
Support different protocols of fetching stubs / contracts


marcingrzejszczak:
## Introduction 

We already support pluggable Stub Downloading via the `StubDownloaderBuilder` (https://cloud.spring.io/spring-cloud-static/Edgware.SR2/multi/multi__using_the_pluggable_architecture.html#_using_the_custom_stub_downloader). That means that if someone wants to download stubs / contracts from e.g. S3 or git then it's possible to be done.

The `StubDownloaderBuilder` interface, that is responsible for building a `StubDownloader` looks like this

```java
public interface StubDownloaderBuilder {

	StubDownloader build(StubRunnerOptions stubRunnerOptions);
}
```

Let's take a look at `StubDownloader`:

```java
public interface StubDownloader {

	/**
	 * Returns a mapping of updated StubConfiguration (it will contain the resolved version) and the location of the downloaded JAR.
	 * If there was no artifact this method will return {@code null}.
	 */
	Map.Entry<StubConfiguration,File> downloadAndUnpackStubJar(StubConfiguration stubConfiguration);
}
```

The interface returns a folder with contracts and stubs for the given coordinates of a stub. 

## The Problem

Currently, you can have only one way of fetching contracts / stubs at a time. It would be great to easily allow plugging in some new extension mechanisms without interrupting the current flow. Also we could add some MVP for 2.0.0 to allow some other mechanisms. E.g. `git` and `s3`.

## Proposed Solution

If `StubDownloaderBuilder` had another method like `boolean isAccepted(StubRunnerOptions options)` then we could iterate over all builders and only build a `StubDownloader` for these components that can accept it. What would the logic be like? For example, for `S3StubDownloader` we would have sth like this:

```java
boolean isAccepted(StubRunnerOptions options) {
    return options.getRepositoryRoot().startsWith("s3://");
}
```

The current method `downloadAndUnpackStubJar` would get renamed to sth like `contractsAndStubsLocation()` (please comment your ideas).

```java
Map.Entry<StubConfiguration,File> contractsAndStubsLocation(StubConfiguration stubConfiguration);
```

The implementations of the builder and stub downloader would need to have some properties passed. Remember, that the builder is used both within the plugins and the apps (not necessarily those that have Spring Context setup). The implementations could try reading the properties (like usernames & passwords) from env variables, but sometimes it's not enough. That's why we could allow passing of custom, additional properties to the plugins. E.g.

.Maven

```xml
<plugin>
[...]

PR #593, opened 2018-03-22
Iterate over stub runner downloader builders

marcingrzejszczak:
with this change we try to retrieve a list of builders of stub runner downloaders. The idea is that if a builder can't build a stub runner downloader, it returns null. If a stub runner downloader can't download an artifact it returns null. That way we can iterate over builders and try to point to unpacked stubs when applicable. No longer do we pick only one applicatble builder. We pick the first matching one.

first attempt to work with #580

### DIFF
--- a/spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/AetherStubDownloaderBuilder.java
+++ b/spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/AetherStubDownloaderBuilder.java
@@ -1,0 +1,21 @@
+package org.springframework.cloud.contract.stubrunner;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.cloud.contract.stubrunner.spring.StubRunnerProperties;
+
+/**
+ * @author Marcin Grzejszczak
+ * @since 2.0.0
+ */
+public class AetherStubDownloaderBuilder implements StubDownloaderBuilder {
+	private static final Log log = LogFactory.getLog(AetherStubDownloaderBuilder.class);
+
+	@Override public StubDownloader build(StubRunnerOptions stubRunnerOptions) {
+		if (stubRunnerOptions.stubsMode == StubRunnerProperties.StubsMode.CLASSPATH) {
+			return null;
+		}
+		log.info("Will download stubs from classpath");
+		return new AetherStubDownloader(stubRunnerOptions);
+	}
+}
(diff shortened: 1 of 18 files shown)


spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/AetherStubDownloaderBuilder.java
```diff
@@ -0,0 +1,21 @@
...
+ * @since 2.0.0
+ */
+public class AetherStubDownloaderBuilder implements StubDownloaderBuilder {
+	private static final Log log = LogFactory.getLog(AetherStubDownloaderBuilder.class);
+
+	@Override public StubDownloader build(StubRunnerOptions stubRunnerOptions) {
+		if (stubRunnerOptions.stubsMode == StubRunnerProperties.StubsMode.CLASSPATH) {
+			return null;
+		}
+		log.info("Will download stubs from classpath");
```
dstepanov: Wrong log

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ClasspathStubProvider.java
```diff
@@ -53,14 +53,18 @@
...
-			.getLog(MethodHandles.lookup().lookupClass());
+			.getLog(ClasspathStubProvider.class);
 
 	private static final int TEMP_DIR_ATTEMPTS = 10000;
 	private final PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver(
 			new DefaultResourceLoader());
 
 	@Override
 	public StubDownloader build(final StubRunnerOptions stubRunnerOptions) {
+		if (stubRunnerOptions.stubsMode != StubRunnerProperties.StubsMode.CLASSPATH) {
```
dstepanov: I would expect `== StubsMode.REMOTE`

marcingrzejszczak: You can have `StubsMode.LOCAL`

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/AetherStubDownloaderBuilder.java
marcingrzejszczak: 😱😱😱 great catch!

marcingrzejszczak:
Merged via https://github.com/spring-cloud/spring-cloud-contract/commit/01e682044c0f0a2c36b86181966b1afd1cf97aa7

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/AetherStubDownloaderBuilder.java
dstepanov: I expect mention in the release notes about this great find! :laughing:

PR #624, opened 2018-04-21
Avoid body parsing for HEAD request

axelhodler:
Maybe this is behaviour more users would find helpful.

We had an issue with a failing build due to FindBugs running over the generated verifier tests (now added to the excluded packages).

In the case of `HEAD` the `responseAsString` was not used and FindBugs cried "unused variable" and failed.

Of course it could be improved upon. Something along the lines of ```IF assert on response THEN parse response```

But IMHO this would go against the spec for [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD)

### DIFF
--- a/spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/JaxRsClientJUnitMethodBodyBuilder.groovy
+++ b/spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/JaxRsClientJUnitMethodBodyBuilder.groovy
@@ -73,7 +73,9 @@ class JaxRsClientJUnitMethodBodyBuilder extends JUnitMethodBodyBuilder {
 		bb.unindent()
 
 		bb.addEmptyLine()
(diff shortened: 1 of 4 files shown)


marcingrzejszczak:
Hmm i think that if your don't specify the response body in the contract  then this line shouldn't be printed regardless of the http method

axelhodler:
> Hmm i think that if your don't specify the response body in the contract then this line shouldn't be printed regardless of the http method

yes, a lot better!

axelhodler:
Rebased and applied the improvement idea.

If the approach is fine it could be applied to the other `BodyBuilders` too.

marcingrzejszczak:
Perfect! The approach is great, let's continue with the rest and merge the PR :)

axelhodler:
functionality now includes `JaxRsClientSpockMethodRequestProcessingBodyBuilder`

`MockMvcSpockMethodBuilder` and `MockMvcJUnitMethodBuilder` already only parse the body if we assert on it 👍

marcingrzejszczak:
Awesome work @axelhodler , as usual :)

PR #634, opened 2018-05-02
snapshotCheckSkip javadoc enhancement

yuranos:
@marcingrzejszczak, I will start small:)

Do you think it can be useful? Otherwise it can be confusing because stubs jar will still be in the local repo even when Git is used.


marcingrzejszczak:
Thanks :)

yuranos:
@marcingrzejszczak , I noticed you are using plain string manipulations instead of something like codeModel or JavaPoet. Are you considering them as options for future improvements?

marcingrzejszczak:
AFAIR we did some analysis once about tools like that and there was no real value for our particular use case. But feel free to create a small spike on this.

PR #641, opened 2018-05-06
Clarifying behaviour of Aether artifact downloader for transitive dependencies

yuranos:
@marcingrzejszczak , my suggestion here might be wrong, but I was not able to reproduce transitive dependencies loading and wanted to grab your attention to this peace of documentation.

My steps: I'm adding some absolutely random dependencies to my producer. I build a producer with its stubs. Afterwards I clean-up .m2.
Naturally, the next step is to build a consumer. And when I do so, I don't find the dependencies needed for producer downloaded into .m2.

I'm not an expert in Aether, but it will probably require a bit more logic to download transitive dependencies, like in this [example](http://wiki.eclipse.org/Aether/Resolving_Dependencies). And, luckily, I don't see any code akin to this anywhere in `spring-cloud-contract`.

This part of documentation can do with a bit of improvement anyway. Let me know if my suggestion is wrong, so that I can at least refactor what is already there.

### DIFF
--- a/docs/src/main/asciidoc/verifier_setup.adoc
+++ b/docs/src/main/asciidoc/verifier_setup.adoc
@@ -902,7 +902,7 @@ In order to fix this issue, provide the following section in your `pom.xml`:
 
 === Stubs and Transitive Dependencies
 
-The Maven and Gradle plugin that add the tasks that create the stubs jar for you. One
+The Maven or Gradle plugin should be used to add the tasks that create the stubs jar for you. One
 problem that arises is that, when reusing the stubs, you can mistakenly import all of
 that stub's dependencies. When building a Maven artifact, even though you have a couple
 of different jars, all of them share one pom:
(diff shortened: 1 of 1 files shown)


marcingrzejszczak:
I'm sorry but I completely don't understand the problem and how this documentation update should help. Isn't it quite obvious that if you want to fetch the dependencies they need to be first uploaded somewhere? Either locally or in a remote location?

yuranos:
Hi @marcingrzejszczak .
My suggestion can be wrong, of course. The reason I created this PR is because I was not able to reproduce the scenario with TRANSITIVITY. Let's say I add a copletely irrelevant dependency to my producer. Let it be bountcycastle. After I `mvn install` the producer with its stubs I delete bouncycastle from .m2.
Then I switch to the consumer. And no matter what I do in it, it will never download bouncycastle back to .m2 if I don't add `stubs` as a dependency, but simply use stub-runner.

marcingrzejszczak:
Can you create a small, reproducible sample of this issue?

PR #644, opened 2018-05-07
Matchers documentation improvements

yuranos:
Stubs matchers - always about REQUESTS;
Verifier matchers - always about RESPONSES.

### DIFF
--- a/docs/src/main/asciidoc/verifier_contract.adoc
+++ b/docs/src/main/asciidoc/verifier_contract.adoc
@@ -764,34 +764,34 @@ following matching possibilities:
 
 .Groovy DSL
 
-* For the stubs:
-** `byEquality()`: The value taken from the response via the provided JSON Path must be
+* For the stubs(in tests on the Consumer's side):
+** `byEquality()`: The value taken from the consumer's request via the provided JSON Path must be
 equal to the value provided in the contract.
-** `byRegex(...)`: The value taken from the response via the provided JSON Path must
+** `byRegex(...)`: The value taken from the consumer's request via the provided JSON Path must
 match the regex.
-** `byDate()`: The value taken from the response via the provided JSON Path must
+** `byDate()`: The value taken from the consumer's request via the provided JSON Path must
 match the regex for an ISO Date value.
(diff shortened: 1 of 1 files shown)


marcingrzejszczak:
Shouldn't you update the YAML section too? Or is it correct?

yuranos:
YAML is fine because it doesn't contain a detailed explanation:)

yuranos:
@marcingrzejszczak , do you have Slack channel or something like that to ask some questions? I'm not sure SO will fit to that purpose because the questions are sometimes about approaches and not necessarily technical.

For example, in [here](https://cloud.spring.io/spring-cloud-contract/single/spring-cloud-contract.html#_pact_consumer_producer_contract_approach).
Who will publish Pact contract if it's not Consumer? How can Producer do it via spring-cloud-contact toolkit?
What's more, the stubs will not be published to PACT Broker, I guess. That means we will have contacts and stubs spread out between stub jar and Pact Broker. Can be a bit confusing.

marcingrzejszczak:
You can chat here - https://gitter.im/spring-cloud/spring-cloud-contract

> Who will publish Pact contract if it's not Consumer? How can Producer do it via spring-cloud-contact toolkit?

Producer can reuse the code that the consumers use, to publish pact files to the broker. From the pact files you can also generate tests on the producer side to assert if the producer is telling the truth.

> What's more, the stubs will not be published to PACT Broker, I guess.

They won't but we are capable of converting the downloaded pact files to stubs - check this out: https://github.com/spring-cloud-samples/spring-cloud-contract-samples/tree/2.0.x/consumer_pact_stubrunner

> That means we will have contacts and stubs spread out between stub jar and Pact Broker. Can be a bit confusing.

This is an example of doing provider contracts with a tool that doesn't support it out of the box. That's why it's confusing.

yuranos:
Thanks, @marcingrzejszczak 
So, if I understood `PactStubDownloaderBuilder` correctly we generate stubs independently on both sides?

marcingrzejszczak:
Yeah, exactly. It has to since it's violating the principles of the consumer contract approach of Pact.

PR #675, opened 2018-06-30
Improve docs on contract priority

axelhodler:
Ultra minor change.

The wording of `greater priority` led to an issue in a project where a contract with `priority(999)` (supposedly match first) and `priority(888)` (supposedly match second) was created. The example how `priority(1)` would match before `priority(2)` would have resolved this I guess :)


marcingrzejszczak:
I thought that I have already fixed that part but apparently not. Thanks a lot!!!!

PR #690, opened 2018-07-15
Add junit5 support - fixes #489

OlgaMaciaszek:
Fixes https://github.com/spring-cloud/spring-cloud-contract/issues/489

PR #692, opened 2018-07-16
Fix using regex patterns when matching raw body value

alek-sys:
When matching non-json or xml body with regex the expression gets
escaped. Because of that all standard patterns (number(), positiveInt())
will not work as well as any other regex involving escaped characters.

E.g.
`response {
  status OK()
  body(value(stub("1"), test(number())))
}`

Will generate invalid matcher:

`assertThat(responseBody).matches("-?(d*.d+|d+)");`

Valid version is:

`assertThat(responseBody).matches("-?(\\d*.\\d+|\\d+)");`

### DIFF
--- a/spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/JUnitMethodBodyBuilder.groovy
+++ b/spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/JUnitMethodBodyBuilder.groovy
@@ -26,12 +26,14 @@ import org.springframework.cloud.contract.spec.internal.Header
 import org.springframework.cloud.contract.spec.internal.NamedProperty
 import org.springframework.cloud.contract.spec.internal.Request
 import org.springframework.cloud.contract.verifier.config.ContractVerifierConfigProperties
+import org.springframework.cloud.contract.verifier.util.RegexpBuilders
 
 import java.util.regex.Pattern
 
 import static groovy.json.StringEscapeUtils.escapeJava
 import static org.springframework.cloud.contract.verifier.config.TestFramework.JUNIT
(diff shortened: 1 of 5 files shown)


alek-sys:
`createBodyComparison` methods in `SpockMethodRequestProcessingBodyBuilder.groovy` and `JUnitMethodBodyBuilder.groovy` as basically the same, but generate different code depending on framework (Spock, JUnit). I was thinking to push it to the base class, but it feels to me that here a bit of copy-paste is better that abstraction.

marcingrzejszczak:
First we'll have to merge https://github.com/spring-cloud/spring-cloud-contract/pull/690 . I'll skip reviewing this PR until I merge that one. First, we'll need to branch the current master to `2.0.x` branch and start a `2.1.0` branch.

alek-sys:
Cool, I'll keep an eye on this one and update all the tests to JUnit 5 if needed.

spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/JUnitMethodBodyBuilder.groovy
```diff
@@ -20,18 +20,16 @@ import groovy.json.StringEscapeUtils
 import groovy.transform.PackageScope
 import groovy.transform.TypeChecked
 import org.springframework.cloud.contract.spec.Contract
-import org.springframework.cloud.contract.spec.internal.Cookie
-import org.springframework.cloud.contract.spec.internal.ExecutionProperty
-import org.springframework.cloud.contract.spec.internal.Header
-import org.springframework.cloud.contract.spec.internal.NamedProperty
-import org.springframework.cloud.contract.spec.internal.Request
+import org.springframework.cloud.contract.spec.internal.*
```
marcingrzejszczak: Can we not group the imports please?

marcingrzejszczak:
Congrats @alek-sys on a great PR :)

PR #757, opened 2018-10-22
Webtestclient support

OlgaMaciaszek:


spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/WebTestClientJUnitMethodBodyBuilder.groovy
```diff
@@ -0,0 +1,47 @@
...
+@PackageScope
+class WebTestClientJUnitMethodBodyBuilder extends RestAssuredJUnitMethodBodyBuilder {
+
+    WebTestClientJUnitMethodBodyBuilder(Contract stubDefinition, ContractVerifierConfigProperties configProperties) {
+        super(stubDefinition, configProperties)
+    }
+
+    @Override
+    protected String returnedResponseType() {
+        return 'WebTestClientResponse'
```
marcingrzejszczak: Why, oh why do you use single `'` ? :( :( :(

spring-cloud-contract-verifier/src/test/groovy/org/springframework/cloud/contract/verifier/util/SyntaxChecker.groovy
```diff
@@ -106,15 +118,22 @@ class SyntaxChecker {
...
 		sourceCode.append(test)
 		return new GroovyShell(SyntaxChecker.classLoader, configuration).parse(sourceCode.toString())
 	}
 
-	static Class tryToCompileJava(String test) {
+	private static GString getStaticImports(String builderName) {
+		if (builderName.toLowerCase().contains('webtestclient')) {
+			return "$WEB_TEST_CLIENT_STATIC_IMPORTS\n"
+		}
+		"$STATIC_IMPORTS\n"
```
marcingrzejszczak: `return` is missing and the method should return `String`

spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/WebTestClientJUnitMethodBodyBuilder.groovy
OlgaMaciaszek: Cause I like it.

OlgaMaciaszek: And cause we had a crazy mixture of both.

PR #780, opened 2018-11-05
Improved spring-cloud-wiremock documentation

ioanngolovko:
Sometimes people want to use methods like `addMockServiceRequestListener`, which are not static. But there is no any notes about  `WireMockServer` in spring-cloud-wiremock documentation. 

I saw how the developers did not know about it and made a mistake. 
They mixed junit4 and junit5 even for new services, because `WireMockRule` inherits from `WireMockServer` and  provides non static methods.

This situation may seem funny, but many can get into it.


marcingrzejszczak:
Congrats @ioanngolovko on your first contribution :)

PR #782, opened 2018-11-06
Update README.adoc Mentioning both Maven and Gradle plugin testFramework and targetFramework property

jntakpe:
Fix wrong property name to change the test framework.

### DIFF
--- a/docs/src/main/asciidoc/verifier_introduction.adoc
+++ b/docs/src/main/asciidoc/verifier_introduction.adoc
@@ -395,8 +395,8 @@ Here is an example of a test generated in `WEBTESTCLIENT` test mode:
 	}
 ----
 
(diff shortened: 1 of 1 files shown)


marcingrzejszczak:
The documentation is generated from the docs module. Can you update the files there?

jntakpe:
Yes of course

jntakpe:
@marcingrzejszczak Updated actually testFramework is used by Maven plugin and targetFramework by Gradle

OlgaMaciaszek:
Yes, @jntakpe the two different names were a bug - it's been fixed here: https://github.com/spring-cloud/spring-cloud-contract/commit/8c773868160ed6000701c154f95b02c79b6a10e5 
so now both will have the right name.

Issue #803, opened 2018-11-26
Cloud Stub Runner Boot Application and Cloud Stream (Kafka) for Smoke Test doesn't work


DanceSchoolMicroservices:
In August i opened issue in  [stack overflow](https://stackoverflow.com/questions/51731984/spring-cloud-stub-runner-boot-application-and-cloud-stream-kafka-for-smoke-tes). I have upgraded to the Finchley.SR2 and have the same problem. I have also seen that github-analytics-stub-runner-boot sample has been upgraded to the Finchley.SR1 and I use the same configuration that in the sample. I think the problem is still StreamsRelatedAutoConfiguration. It could be nice set `@ConditionalOnProperty(
    name = {"stubrunner.stream.test.enabled"},
    havingValue = "true",
    matchIfMissing = true
)` to excude TestSupportBinderAutoConfiguration
Thanks

marcingrzejszczak:
Can you file a pull request?

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

spring-issuemaster:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

PR #805, opened 2018-11-28
fix stub runner application when is in real stream mode

davsuapas:
Do reference to issue: #803
I have cheched it and it works using `@SpringBootApplication( exclude = { TestSupportBinderAutoConfiguration.class })`


OlgaMaciaszek:
Thanks for the PR. Looks good.

OlgaMaciaszek:
Fixes gh-803

PR #813, opened 2018-12-11
Upgrade and parameterize Maven dependencies

dsyer:


marcingrzejszczak:
I think I didn't upgrade maven cause takari plugin required that version (but maybe I'm wrong).

marcingrzejszczak:
What do we do about this @dsyer ?

dsyer:
I guess we should do it, if nothing breaks. I have some issues with the thin launcher IIRC, but I probably need to sort those out separately.

Issue #837, opened 2019-01-03
Servlet API in wiremock causes Boot to activate web app in stub runner app


dsyer:
This change made `spring-boot-starter-web` optional: https://github.com/spring-cloud/spring-cloud-contract/blame/34f7cd31edae750a9dcd3a7b5493cd19129c3383/spring-cloud-contract-starters/spring-cloud-starter-contract-stub-runner/pom.xml. It makes sense because the user app might be webflux (or even not a web app).

The problem is that the presence of the Servlet API triggers Spring Boot to think it needs an embedded webserver, and one might not be available now.

We could perhaps solve the problem by adding an `EnvironmentPostProcessor`. If wiremock is there and spring-webmvc is not, set `spring.main.web-application-type=<something>`. I guess `<something>` has to depend on whether or not webflux/netty is present.

marcingrzejszczak:
That's actually no longer the case. Web does not leak out from stub runner modules

PR #839, opened 2019-01-04
Sets web-application-type depending on presence of WebMvc / WebFlux

marcingrzejszczak:
without this change, since WireMock is using the Servlet API, Spring Boot falsely assumes that even a reactive application is a servlet one
with this change we first analyze whether WebMvc related classes are present and then we set the web application type

fixes gh-837


dsyer:
LGTM. I’d be interested to hear if the Boot team has anything to add.

marcingrzejszczak:
cc @snicoll @wilkinsona @philwebb ?

snicoll:
> without this change, since WireMock is using the Servlet API, Spring Boot falsely assumes that even a reactive application is a servlet one

I am not sure I got that. Can you share the auto-config report that you get (i.e. without this change)

marcingrzejszczak:
That's not required since we've managed to make the web dependencies optional for stub runner

PR #870, opened 2019-02-04
Set default handler for XmlSlurper to avoid printing exception

konstantinshevchuk:
[Fatal Error] :1:1: Content is not allowed in prolog


konstantinshevchuk:
@ryanjbaxter I create a new branch against `2.0.x`. Unfortunately I can't build a project locally against 2.0x. :( Hopefully remotely it's not going to fail ;)

Issue #883, opened 2019-02-13
Wiremock: Spring Context is always marked as dirty even when port is 0.


boothen:
With version 2.1.0.RELEASE the Spring context is always marked as dirty after each test run. This causes tests to run much slower as the Spring context is initialised each time.

The issue is in the following code in WireMockTestExecutionListener

```
	private boolean portIsFixed(TestContext testContext) {
		WireMockConfiguration wireMockProperties = wireMockConfig(testContext);
		int httpPort = wireMockProperties.wireMock.getServer().getPort();
		int httpsPort = wireMockProperties.wireMock.getServer().getHttpsPort();
		return (httpPort != 0 || httpsPort != -1) && httpsPort != 0;
	}
```

The httpPort is either set to a fixed value or given a dynamic value in WireMockApplicationListener. Therefore, httpPort can never be 0.

I've submitted a PR to workaround this issue, basically the code needs to determine if the port is dynamic based off the property value and not the httpPort value.

I've included a screenshot that runs the same test twice. The output from the 2nd test shows the "You've used fixed ports ..." warning message. 

<img width="1166" alt="screenshot 2019-02-13 at 23 07 50" src="https://user-images.githubusercontent.com/1582761/52750472-3cc36c80-2fe4-11e9-9e55-7815d65dc4a2.png">

marcingrzejszczak:
Should be fixed via https://github.com/spring-cloud/spring-cloud-contract/commit/11965323ae947775927adb9564c072b4f848e28e

boothen:
With 2.1.1.RELEASE I've noticed that the issue still occurs. The fix works for JUnit/SpringRunner tests. However, when using the framework with Cucumber it turns out that the context is refreshed multiple times.

On context initialisation the logic sets the portDynamic and httpsPortDynamic to true, but on subsequent refreshes the logic updates httpsPortDynamic to false.

I've submitted a Pull Request with the fix. However, I'm unable to provide a test to verify that the issue is fixed in the context refresh scenario.

OlgaMaciaszek:
Reopening issue since PR was causing exceptions. Please provide a fixed PR.

OlgaMaciaszek:
Reopening issue since PR was causing exceptions. Please provide a fixed PR.

OlgaMaciaszek:
This should be fixed by https://github.com/spring-cloud/spring-cloud-contract/pull/1048

PR #884, opened 2019-02-13
#883: Wiremock: Fix for Spring context always being set to Dirty.

boothen:
This is a fix for #883.

### DIFF
--- a/spring-cloud-contract-wiremock/src/test/java/org/springframework/cloud/contract/wiremock/AutoConfigureWireMockRandomPortApplicationTests.java
+++ b/spring-cloud-contract-wiremock/src/test/java/org/springframework/cloud/contract/wiremock/AutoConfigureWireMockRandomPortApplicationTests.java
@@ -49,7 +49,7 @@ public class AutoConfigureWireMockRandomPortApplicationTests {
 	}
 
 	@Test
-	public void contextIsNotReload() throws Exception {
+	public void dynamicPortsAreResolvedCorrectly() throws Exception {
 		assertThat(wireMockProperties.getServer().getPort()).isNotEqualTo(0);
 		assertThat(wireMockProperties.getServer().isPortDynamic()).isTrue();
 		assertThat(wireMockProperties.getServer().isHttpsPortDynamic()).isTrue();
--- a/spring-cloud-contract-wiremock/src/test/java/org/springframework/cloud/contract/wiremock/AutoConfigureWireMockRandomPortApplicationTests.java
+++ b/spring-cloud-contract-wiremock/src/test/java/org/springframework/cloud/contract/wiremock/AutoConfigureWireMockRandomPortApplicationTests.java
@@ -38,11 +38,21 @@ public class AutoConfigureWireMockRandomPortApplicationTests {
 	@Autowired
 	private Service service;
 
+	@Autowired
+	private WireMockProperties wireMockProperties;
+
 	@Test
 	public void contextLoads() throws Exception {
 		stubFor(get(urlEqualTo("/test")).willReturn(aResponse()
 				.withHeader("Content-Type", "text/plain").withBody("Hello World!")));
 		assertThat(this.service.go()).isEqualTo("Hello World!");
+	}
+
+	@Test
+	public void contextIsNotReload() throws Exception {
+		assertThat(wireMockProperties.getServer().getPort()).isNotEqualTo(0);
+		assertThat(wireMockProperties.getServer().isPortDynamic()).isTrue();
+		assertThat(wireMockProperties.getServer().isHttpsPortDynamic()).isTrue();
 	}
 
 }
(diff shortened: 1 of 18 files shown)


pivotal-issuemaster:
@boothen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-contract&pullRequestId=884)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-contract&pullRequestId=884) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@boothen Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-contract&pullRequestId=884)!

spring-cloud-contract-wiremock/src/test/java/org/springframework/cloud/contract/wiremock/AutoConfigureWireMockRandomPortApplicationTests.java
```diff
@@ -38,11 +38,21 @@
...
+
 	@Test
 	public void contextLoads() throws Exception {
 		stubFor(get(urlEqualTo("/test")).willReturn(aResponse()
 				.withHeader("Content-Type", "text/plain").withBody("Hello World!")));
 		assertThat(this.service.go()).isEqualTo("Hello World!");
 	}
 
+	@Test
+	public void contextIsNotReload() throws Exception {
```
OlgaMaciaszek: Maybe change test method name to something like `dynamicPortsAreResolvedCorrectly`? You don't really assert here if the context is reloaded.

boothen: I've updated the code as directed. Thanks for coming back to me so quick.

OlgaMaciaszek: Sure. However, I see the build is failing exactly with the issue that checking ports was introduced to fix, so it looks like, there has been no context cleanup where it needed to happen.

boothen: Sorry about that. Late night code changes without applicable tests never workout well. I've fixed the logic around portIsFixed.

The tests were failing as previous classes weren't being marked to have the Spring context reloaded. The code worked before the change as it always reloaded the Spring context. I don't think it's a good idea that a test can pass in isolation but fail when run as part of a suite. Therefore, I've added a new test the applicable test classes to verify that the portIsFixed or not fixed depending on the configuration. I hope that's ok?

OlgaMaciaszek:
@boothen Thanks for fixing the tests. I also think it's a good idea to have somewhere a test that verifies fixed ports and a test that verifies not fixed ports. But I don't think we should add more of these same tests to different classes, as they verify a single code execution path - the behaviour of ports does not depend on other conditions set in different tests, so please leave one of each (portFixed, porstNotFixed) and remove the rest. That should be the last change :), and then we can merge.

boothen:
Removed most of the portIsFixed/portNotFixed tests. I've left in 4 which cover different port configurations. i.e port=0, port=9999, port = 0 & httpsPort=9999, port = 0 and httpsPort =0.

marcingrzejszczak:
Hey @boothen have you rebase against master? Cause I see a couple of checkstyle issues that should have broken your build.

boothen:
Sorry about that. I had the wrong codestyle configured. Should be ok now?

OlgaMaciaszek:
Hello @boothen Some tests have started failing again. Have you tried merging against master and then rebuilding this locally? Does it pass locally?

OlgaMaciaszek:
Ok, @boothen I see it's passing now. Merging. Thanks for the fix.

boothen:
Hi @OlgaMaciaszek,

I believe there could be random test failures due to the running order of tests. The test classes with random ports won't dirty the Spring Context. This means if 2 test classes are run consecutively that use random ports, the 2nd class would use the same as the 1st. Even if there are differences in the test configuration.

For instance, if AutoConfigureWireMockStubsAndMultipleFilesApplicationTests was executed followed by AutoConfigureWireMockStubsApplicationTests, the context wouldn't be restarted to use the new stub mapping in the 2nd test class and so could fail.

The workaround for this issue would be to add the DirtiesContext annotation to all test classes that use the random ports.

Thanks.
Matt.

marcingrzejszczak:
> The workaround for this issue would be to add the DirtiesContext annotation to all test classes that use the random ports.

This is the very essence of this PR. Not to do this. If a user does the same in their codebase, will we ask them to use `DirtiesContext` ?

boothen:
The issue fixed is that the context was always being set to dirty whether dynamic or fixed ports were used. With the fix in place dynamic ports now won't reload the context.

However, some automated tests with dynamic ports use different wiremock configuration as part of their tests, so need the context reloading. e.g. files or stub configuration are different.

A user in their codebase would more likely be using the same wiremock configuration across their tests, so less likely to need to use the DirtiesContext.

PR #1028, opened 2019-03-28
#883: Further DirtiesContext fix where the context is refreshed during the test phase.

boothen:
The original commit fixed the DirtiesContext issue for JUnit/SpringRunner tests. However, the issue was not fixed where the context was refreshed during the test phase. This occurs when using test frameworks like cucumber.

This latest PR avoids updating the httpsPortDynamic where a fixed port has been specified.

PR #1040, opened 2019-04-09
#1028: Register port now only handles auto port values.

boothen:
This is to avoid the httpsPortDynamic being set to true on the context initialisation, but then being updated to false when the context is refreshed.

Fix for #1028

### DIFF
--- a/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockApplicationListener.java
+++ b/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockApplicationListener.java
@@ -56,8 +56,9 @@ public class WireMockApplicationListener
 					SocketUtils.findAvailableTcpPort(10000, 12500));
 			source.put("wiremock.server.port-dynamic", true);
 		}
-		if (environment.getProperty("wiremock.server.https-port", Integer.class,
-				0) == 0) {
+		int httpsPortProperty = environment.getProperty("wiremock.server.https-port",
+				Integer.class, 0);
+		if (httpsPortProperty == 0) {
 			MutablePropertySources propertySources = environment.getPropertySources();
 			addPropertySource(propertySources);
 			Map<String, Object> source = ((MapPropertySource) propertySources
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
Hi @boothen , thanks for the PR! Any chances for a test for this?

boothen:
Hi @marcingrzejszczak, the original PR contained a test for the fix. However, this PR is to resolve an issue where the test framework refreshes the Spring context. The refresh caused the original code to update the httpsPortDynamic from true to false. 

Cucumber refreshes the Spring context after each test scenario. However, I'm struggling to trigger a refresh using JUnit. Using the DirtiesContext annotation doesn't work the same way.

Any suggestions on how I could recreate it?

marcingrzejszczak:
I've merged it, it might be tricky to recreate it

marcingrzejszczak:
I'll have to revert this change cause it broke our samples

marcingrzejszczak:
You can try to do the change again, but be sure you haven't broken the https://github.com/spring-cloud-samples/sc-contract-car-rental sample

boothen:
Hi @marcingrzejszczak , could you elaborate further? 

Is it a JUnit test that is failing? I can only see one test that uses the AutoConfigureWireMock and that has a fixed port so will be unaffected by the code change.

I've successfully built the sample project against a local 2.2.0-SNAPSHOT, so not sure what the issue is.

Any help is appreciated, without the fix we're unable to use the latest version.

marcingrzejszczak:
Everything works fine right now cause i reverted your commit.

boothen:
Not sure I follow. I built the sample project with a local build of spring-cloud-contract that included my changes. I was expecting to see a test failure when building the sample project this way.

What is the breakage you're seeing in the sample project or how can I reproduce it? Unless I can reproduce it there's no way I can understand how it was broken.

boothen:
Think I've managed to reproduce it. Were you seeing

> java.net.BindException: Address already in use

during the tests?

OlgaMaciaszek:
Hi @boothen - yes, that's exactly the exception we were getting.

```
[ERROR] should_retrieve_list_of_frauds_from_stub(com.example.carrental.RentACarTests)  Time elapsed: 0.06 s  <<< ERROR!

com.github.tomakehurst.wiremock.common.FatalStartupException: java.lang.RuntimeException: java.net.BindException: Address already in use

Caused by: java.lang.RuntimeException: java.net.BindException: Address already in use

Caused by: java.net.BindException: Address already in use
```

boothen:
Hi there,

I've identified a further fix for the issue seen in the sample project. It looks like when the RentACarTest context is initially loaded it hasn't scanned the annotation properties. I think this is due to the StubRunnerRule in the test. Removing the StubRunnerRule the 

> Caused by: java.net.BindException: Address already in use

error doesn't occur.

I can submit a PR with a fix for this as the annotation properties are scanned on subsequent context refresh. However, again I'm struggling to write a unit test to prove it. I could possibly add a test to the spring-cloud-contract-stub-runner to verify it, but it looks like not all tests are being executed in that project. #1047

Should I submit the PR without a test while the test execution issue is investigated?

boothen:
I've submitted PR #1048 for this. It doesn't have a test case with it unfortunately. I'll try to add one when/if #1047 is resolved.

Issue #1088, opened 2019-05-28
Hypermedia-aware stubs


dsyer:
This might be more of a Wiremock feature (dunno), but I figure it would be useful. When a service is hypermedia capable, it will return links, e.g.

```
GET /

{"_links":{"markup":{"href":"http://localhost:8080/documents"},"guides":{"href":"http://localhost:8080/guides/"}}}
```

If you create stubs from this output it has port 8080 hardcoded, so when the stub runner starts up on a different port the links will be wrong. It would be neat to be able to map the port, or do a placeholder replacement or something.

Maybe this is helpful: http://wiremock.org/docs/response-templating/. Looks like maybe a stub defined as 

```
{"_links":{"markup":{"href":"http://localhost:{{request.port}}/documents"},"guides":{"href":"http://localhost:{{request.port}}/guides/"}}}
```

might work, if the template transformer is applied (which we could do declaratively somehow in Spring Cloud)?

dsyer:
This works:

```java
@RunWith(SpringRunner.class)
@SpringBootTest
public class DemoApplicationTests {
	
	private RestTemplate rest = new RestTemplate();

	@ClassRule
	public static WireMockClassRule wiremock = new WireMockClassRule(
			WireMockSpring.options().extensions(new ResponseTemplateTransformer(false)).dynamicPort());

	@Test
	public void contextLoads() {
		wiremock.stubFor(get(urlEqualTo("/resource")).willReturn(aResponse().withTransformers("response-template")
				.withHeader("Content-Type", "text/plain").withBody("Hello World {{request.requestLine.port}}!")));
		ResponseEntity<String> entity = rest.getForEntity("http://localhost:" + wiremock.port() + "/resource", String.class);
		System.err.println(entity.getBody());
	}

}
```

So I think what we would need would be an option to a) enable the `ResponseTemplateTransformer` in the stub runner, b) filter links in stubs when they are generated, replacing `localhost:8080` with `localhost:{{request.requestLine.port}}`, also adding the `"transformers": ["response-template"]` to the JSON of stubs that contain links.

dsyer:
The stub runner actually already has a `spring.factories` mechanism for registering WireMock `Extensions`, but `spring-cloud-wiremock` doesn't (oddly - maybe they should support the same features?). Extensions can also be registered without using the stub runner by using `@AutoConfigureWireMock` and adding a bean of type `WireMockConfigurationCustomizer`.

[...]

PR #1109, opened 2019-06-14
Add support for placeholders in Wiremock stubs

dsyer:
Fixes gh-1088


spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/restdocs/WireMockRestDocsConfiguration.java
```diff
@@ -38,9 +43,42 @@
...
+
+		@Override
+		public OperationRequest preprocess(OperationRequest request) {
+			return request;
+		}
+
+		@Override
+		public OperationResponse preprocess(OperationResponse response) {
+			String content = response.getContentAsString();
+			if (content.contains("localhost:8080")) {
```
marcingrzejszczak: Will it always be `:8080` ?

dsyer: That's the default for MockMvc, and users almost never change it - I expect we could make it configurable if anyone asks. Actually I didn't try the reactive version, but I expect it's the same. Maybe we need a test.

dsyer:
That test failure looks bogus to me (works on my cloud).

Issue #1106, opened 2019-06-06
Xml contentType and equalToXml body


artemptushkin:
I found that equalToXml converts to wire-mock equalToXml only if there is no specific xml header is placed otherwise converts to xpath matchers.

Example contract
```
request {
		method 'GET'
		url "/users"
		headers {
			header "Content-Type", "customtype/xml"
		}
		body equalToXml(
                 """<foo><name>Artem</name><jobId>1234567890</jobId</foo>"""
                 )
		response {
			status OK()
		}
	}
```
Actual generated stub:
```
...
"bodyPatterns": [
      {
        "matchesXPath": {
          "expression": "/foo/name/text()",
          "equalTo": "Artem"
        }
      },
      {
        "matchesXPath": {
          "expression": "/foo/jobId/text()",
          "equalTo": "1234567890"
        }
      }
    ]
...
```
Expected generated stub:
```
...
"bodyPatterns": [
    {
         "equalToXml":"<foo><name>Artem</name><jobId>1234567890</jobId></foo>"
    }
]
...
```

If remove header `Content-Type` with xml then everything is ok and `equalToXml` at the stub as the [test](https://github.com/spring-cloud/spring-cloud-contract/blob/c81216c460be31c75c40c033c7395bfcf11ccd97/spring-cloud-contract-verifier/src/test/groovy/org/springframework/cloud/contract/verifier/dsl/wiremock/WireMockGroovyDslSpec.groovy) `should use equalToXml` says

I guess the test `should use xml matchers when content type ends with xml` intersects with `should use equalToXml` 

Or please tell me if I am wrong with `equalToXml` and XML content type logic

OlgaMaciaszek:
Thanks  for reporting this @hustleone Will have a look at this and get back to you.

OlgaMaciaszek:
Yes. It's a bug. Will wait for your PR then.

artemptushkin:
There is the same issue with JSON body and json contentType, see PR

PR #1113, opened 2019-06-24
Matching strategy at body and contentType issue

artemptushkin:
Fixed bug with explicit content type and any matching strategy at the body section gh-1106


OlgaMaciaszek:
Fixes gh-1106

PR #1116, opened 2019-06-27
Refactored class and method generation

marcingrzejszczak:
following two event storming sessions with @pilloPL and @OlgaMaciaszek and the Spring Cloud Contract community we've decided to refactor the core modules of Spring Cloud Contract Verifier.

The main idea is that the current `JavaTestGenerator`'s implementation no longer uses complex logic to generate classes and methods.

We've introduced abstractions such as:
- class body builder (to build classes)
- groovy and java class metadata
- imports (for various frameworks)
- default setup for json, xml, restassured, jaxrs
- class annotations for spock, junit and junit5
- field insertion e.g. for messaging

The point of entry is here: 

```java
/**
 * Builds a single test for the given {@link ContractVerifierConfigProperties properties}
 *
 * @since 1.1.0
 */
public class JavaTestGenerator implements SingleTestGenerator {

	private static final Log log = LogFactory.getLog(JavaTestGenerator.class);

	@Override
	public String buildClass(ContractVerifierConfigProperties properties,
			Collection<ContractMetadata> listOfFiles, String className,
			String classPackage, String includedDirectoryRelativePath) {
		throw new UnsupportedOperationException("Deprecated method");
	}

	@Override
	public String buildClass(ContractVerifierConfigProperties properties,
			Collection<ContractMetadata> listOfFiles,
			String includedDirectoryRelativePath, GeneratedClassData generatedClassData) {
		BlockBuilder builder = new BlockBuilder("\t");
		GeneratedClassMetaData metaData = new GeneratedClassMetaData(properties,
				listOfFiles, includedDirectoryRelativePath, generatedClassData);
		return classAsString(builder, metaData);
	}

	private String classAsString(BlockBuilder builder, GeneratedClassMetaData metaData) {
		SingleMethodBuilder methodBuilder = singleMethodBuilder(builder, metaData);
		ClassBodyBuilder bodyBuilder = classBodyBuilder(builder, metaData, methodBuilder);
		GeneratedTestClass generatedTestClass = generatedTestClass(builder, metaData,
				bodyBuilder);
		return generatedTestClass.asClassString();
	}

	GeneratedTestClass generatedTestClass(BlockBuilder builder,
			GeneratedClassMetaData metaData, ClassBodyBuilder bodyBuilder) {
		// @formatter:off
		return GeneratedTestClassBuilder.builder(builder, metaData)
				.classBodyBuilder(bodyBuilder)
					.metaData()
						.java()
						.groovy()
						.build()
					.imports()
						.defaultImports()
						.custom()
						.json()
						.jUnit4()
						.jUnit5()
						.spock()
					.xml()
					.messaging()
					.restAssured()
					.jaxRs()
					.build()
				.classAnnotations()
					.jUnit4()
					.jUnit5()
					.spock()
					.build()
				.build();
		// @formatter:on
	}

	ClassBodyBuilder classBodyBuilder(BlockBuilder builder,
			GeneratedClassMetaData metaData, SingleMethodBuilder methodBuilder) {
		// @formatter:off
		return ClassBodyBuilder.builder(builder, metaData)
				.field()
					.messaging()
					.build()
				.methodBuilder(methodBuilder);
		// @formatter:on
	}

	SingleMethodBuilder singleMethodBuilder(BlockBuilder builder,
			GeneratedClassMetaData metaData) {
		// @formatter:off
		return SingleMethodBuilder.builder(builder, metaData)
				.methodAnnotation()
					.jUnit4()
					.jUnit5()
					.spock()
					.build()
				.methodMetadata()
					.jUnit()
					.spock()
					.build()
				.restAssured()
				.jaxRs()
				.messaging();
		// @formatter:on
	}

	@Override
	public String fileExtension(ContractVerifierConfigProperties properties) {
		return properties.getTestFramework().getClassExtension();
	}

}
```

- [x] refactor the code
- [x] rewrite as much as possible in Java
- [x] update the docs


anatoliy-balakirev:
@marcingrzejszczak I've executed it on our codebase and there seems to be a bug introduced in the `QueryParamsResolver`. For following PACT json:
```javascript
  "interactions": [
    {
      "description": "",
      "request": {
        "method": "GET",
        "path": "/some/path/",
        "query": {
          "user_account_creation_time": [
            null
          ]
        },
        "headers": {
        }
      },
      "response": {
        "status": 400,
        "headers": {
          "Content-Type": "application/problem+json"
        },
        "body": {
          "title": "Bad Request",
          "status": 400,
          "detail": "400 BAD_REQUEST \"Type mismatch.\"; nested exception is org.springframework.beans.TypeMismatchException: Failed to convert value of type 'java.lang.String' to required type 'java.time.ZonedDateTime'; nested exception is org.springframework.core.convert.ConversionFailedException: Failed to convert from type [java.lang.String] to type [@javax.validation.constraints.NotNull @org.springframework.web.bind.annotation.RequestParam @org.springframework.format.annotation.DateTimeFormat java.time.ZonedDateTime] for value 'null'; nested exception is java.lang.IllegalArgumentException: Parse attempt failed for value [null]"
        },
        "matchingRules": {
          "body": {
            "detail": {
              "matchers": [
                {
                  "match": "regex",
                  "regex": ".*Parse attempt failed for value \\[null\\].*\\n?.*"
                }
              ],
              "combine": "AND"
            }
          }
        }
      }
    }
  ]
```
I'm getting null pointer in the `org.springframework.cloud.contract.verifier.builder.QueryParamsResolver`, where you do:
`return value.toString();` and `value` is null.

This seems to do the trick:
`return value == null ? "null" : value.toString();`

marcingrzejszczak:
Argh. Care for a pr?

anatoliy-balakirev:
Will create. Added tests for that interface and found a couple more NPEs there...

In general, to avoid this, I suggest you to start using Spotbugs (successor of the Findbugs): https://spotbugs.github.io/
Or maybe even switch from Java / Groovy to Kotlin? :)

marcingrzejszczak:
> In general, to avoid this, I suggest you to start using Spotbugs (successor of the Findbugs): https://spotbugs.github.io/

Good idea, thanks!

>  Or maybe even switch from Java / Groovy to Kotlin? :)

Oh man, no way :D I'm already rewriting everything from Groovy to Java and it's going to stay that way :)

anatoliy-balakirev:
>Oh man, no way :D I'm already rewriting everything from Groovy to Java and it's going to stay that way :)

For sure I'm not aware of your motivations to do that, but Kotlin is the same concise as Groovy (more less), but more statically typed / feature reach than Java :) E.g. they have a diff between nullable and not nullable types and you can't do `toString()`, for example, on something, which can be null. Compiler will ask you to check it for null first.

There is even auto conversion from java code to kotlin one (was not so good though).

marcingrzejszczak:
> For sure I'm not aware of your motivations to do that, but Kotlin is the same concise as Groovy (more less), but more statically typed / feature reach than Java :) E.g. they have a diff between nullable and not nullable types and you can't do toString(), for example, on something, which can be null. Compiler will ask you to check it for null first.

Cause I believe that the core of the system should be written in Java. There are issues in terms of compatibility between languages. Kotlin and Groovy don't work well with each other but all of them work well with Java. That's why everything should be written in Java :)

anatoliy-balakirev:
Ok, makes sense. 
With Spotbugs it will be a bit painful at the beginning, until you migrate everything, but then it's ok (you'll have to put some asserts in the code though, when property is nullable, but you know it can't be null in this particular place).

There is even more strict checker: https://checkerframework.org/
That's basically kind of an extension to Java's type system and one of the things there is nullability. But I never managed to integrate that. Even for project, which was heavily using Spotbugs it was quite painful.

anatoliy-balakirev:
https://github.com/spring-cloud/spring-cloud-contract/pull/1170

PR #1151, opened 2019-08-02
WIP: Kotlin contract support

TYsewyn:


TYsewyn:
@sdeleuze Would you be able to provide us feedback wrt best practices, code style and such?

sdeleuze:
Hi, thanks for pinging me since after working on MockMVC DSL I wanted to have a deeper look on the Spring Cloud Contract side in order to provide a consistent experience across Spring portfolio. I think what we did on MockMVC DSL can be used as a basis for Spring Cloud Contract, see the related documentation at https://docs.spring.io/autorepo/docs/spring-framework/5.2.0.BUILD-SNAPSHOT/spring-framework-reference/languages.html#mockmvc-dsl. I would advise to create quick tests with Spring Framework 5.2 milestones to play with MockMVC DSL and see what can apply to contracts.

For example, I think it is a best practice to use properties instead functions when possible (even if that's sometimes tricky when multiple values are possible). For example instead of:

```
request {
    url("/foo")
    method("PUT")
    headers {
        header("Content-Language", "fr-FR")
        header("foo", "bar")
    }
}
```

I think it would me more idiomatic and consistent with MockMVC DSL to write:

```
request {
    url = "/foo"
    method = "PUT"
    headers {
        contentLanguage = Locale.FRANCE
        header("foo", "bar")
    }
}
```

`CommonDsl#httpStatus` and `CommonDsl#regexPatterns` look suspicious to me. I am not sure this kind of shortcut is needed, and even if they were they would not be exposed as functions IMO.

DSLs are public APIs so I am not sure we should use the `internal` package.

`ContractDsl.make` does not sound idiomatic, maybe `contract { }` should be used instead?

Any thoughts?

TYsewyn:
Thanks for your input @sdeleuze!
I'm already changing some things to make it more consistent and to let it look like the MockMVC DSL.

Currently I have just one question though.
How would you support multiple types of input for the same property? Something in the lines of the following?
``` kotlin
var method: HttpMethods.HttpMethod? = null

fun method(url: String) {
    val method = when(url) {
        "GET" -> HttpMethods.HttpMethod.GET
        "HEAD" -> HttpMethods.HttpMethod.HEAD
        "POST" -> HttpMethods.HttpMethod.POST
        "PUT" -> HttpMethods.HttpMethod.PUT
        "PATCH" -> HttpMethods.HttpMethod.PATCH
        "DELETE" -> HttpMethods.HttpMethod.DELETE
        "OPTIONS" -> HttpMethods.HttpMethod.OPTIONS
        "TRACE" -> HttpMethods.HttpMethod.TRACE
        else -> throw IllegalArgumentException("Unsupported HTTP method")
    }
}
```

Currently `HttpStatus` and `RegexPatterns` can't be accessed statically, that's why we have these shortcut methods. Without them our DSL would look like eg.:
```kotlin
val httpStatus = HttpStatus()

val contract = ContractDsl.contract {
  ...
  response {
    status = httpStatus.OK()
  }
}
```

But I think we can improve this bit though!

sdeleuze:
For the method, yes something along these lines.

For `HttpStatus`, I guess the class from Spring Cloud Contract is some kind of helper useful in Groovy but I don't think it is a good fit in Kotin or Java (stateless class that requires instantiation to be used). Can you provide me more background on the design of this class? Why not just an enum like we have in Framework?

marcingrzejszczak:
Can you rebase against master, fix the merge conflicts and try to build it again with the new base image for circleci?

specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/MediaTypes.java
```diff
@@ -20,76 +20,248 @@
  * Contains most commonly used media types.
  *
  * @author Marcin Grzejszczak
+ * @author Tim Ysewyn
  * @since 1.0.2
  */
 public class MediaTypes {
```
marcingrzejszczak: shouldn't this be `final` with a `private` constructor?

specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/MessagingHeaders.java
```diff
@@ -20,16 +20,24 @@
  * Contains most commonly used messaging headers.
  *
  * @author Marcin Grzejszczak
+ * @author Tim Ysewyn
  * @since 1.1.2
  */
 public class MessagingHeaders {
```
marcingrzejszczak: shouldn't this be `final` with a `private` constructor?

specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/RegexPatterns.java
```diff
@@ -27,9 +27,12 @@
  * @author Tim Ysewyn
  * @since 1.0.0
  */
-public class RegexPatterns {
+public final class RegexPatterns {
+
+	public RegexPatterns() {
+		System.out.println("WARNING: RegexPatterns shouldn't be instantiated");
```
marcingrzejszczak: `private` constructor and throw an exception upon instantiation instead of doing `sout`

specs/spring-cloud-contract-spec-kotlin/pom.xml
```diff
@@ -0,0 +1,105 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xmlns="http://maven.apache.org/POM/4.0.0"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<properties>
+		<kotlin.version>1.3.41</kotlin.version>
```
marcingrzejszczak: Can we push it to the root `pom.xml` ?

specs/spring-cloud-contract-spec-kotlin/src/main/kotlin/org/springframework/cloud/contract/spec/ContractDsl.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.cloud.contract.spec.internal.Request
+import org.springframework.cloud.contract.spec.internal.RequestDsl
+import org.springframework.cloud.contract.spec.internal.Response
+import org.springframework.cloud.contract.spec.internal.ResponseDsl
+
+/**
+ * @author Tim Ysewyn
+ */
+@ContractDslMarker
+class ContractDsl {
```
marcingrzejszczak: Add `@since` and maybe let's add some javadocs to the methods?

specs/spring-cloud-contract-spec-kotlin/src/main/kotlin/org/springframework/cloud/contract/spec/internal/KotlinContractConverter.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.cloud.contract.spec.Contract
+import org.springframework.cloud.contract.spec.ContractConverter
+import java.io.File
+import java.net.URLClassLoader.newInstance
+import javax.script.ScriptEngineManager
+
+/**
+ * @author Tim Ysewyn
+ */
+class KotlinContractConverter: ContractConverter<List<Contract>> {
```
marcingrzejszczak: Does it make sense to write this in java?

samples/standalone/kotlin/http-client/build.gradle.kts
```diff
@@ -0,0 +1,77 @@
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+	id("org.springframework.boot") version "2.2.0.BUILD-SNAPSHOT"
```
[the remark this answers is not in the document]
TYsewyn: @sdeleuze Can we do that for the plugins too, either in `build.gradle.kts` or `settings.gradle.kts`?  I know we can for the dependencies/management with `val BOM_VERSION: String by project`

specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/MediaTypes.java
TYsewyn: Yes, but that would break backwards compat. We could mark it as depracted and log something to stdout?

marcingrzejszczak: That's a valid point. Let's log a warning then via a logger instead of using sys out.

TYsewyn: Let me do the same like `RegexPatterns`, per your remark.

specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/MessagingHeaders.java
TYsewyn: Let me do the same like `RegexPatterns`, per your remark.

specs/spring-cloud-contract-spec-kotlin/pom.xml
TYsewyn: I thought I did, will fix it.

specs/spring-cloud-contract-spec-java/src/test/groovy/org/springframework/cloud/contract/spec/internal/HttpHeadersTests.java
```diff
@@ -0,0 +1,434 @@
...
+
+package org.springframework.cloud.contract.spec.internal;
+
+import org.assertj.core.api.BDDAssertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * @author Tim Ysewyn
+ */
+public class HttpHeadersTests {
```
OlgaMaciaszek: with Jupiter it should work with a package-private class and test methods. Please verify and, if compiles correctly, adjust all the tests.

specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/HttpStatus.java
```diff
@@ -20,616 +20,1211 @@
  * Helper functions for HTTP statuses.
  *
  * @author Marcin Grzejszczak
+ * @author Tim Ysewyn
  * @since 2.0.0
  */
-public class HttpStatus {
+public final class HttpStatus {
 
 	/**
```
OlgaMaciaszek: Also add private constructor.

specs/spring-cloud-contract-spec-kotlin/src/main/kotlin/org/springframework/cloud/contract/spec/internal/CommonDsl.kt
```diff
@@ -0,0 +1,211 @@
...
+
+import java.io.File
+import java.net.URISyntaxException
+import java.nio.charset.Charset
+import java.util.regex.Pattern
+
+/**
+ * @author Tim Ysewyn
+ */
+open class CommonDsl {
```
OlgaMaciaszek: Is `open` class similar to public scope? if yes, add javadocs with more info (alos description, `@since`, etc.) to all such classes (the public api in general).

specs/spring-cloud-contract-spec-kotlin/src/main/kotlin/org/springframework/cloud/contract/spec/internal/FromRequestDsl.kt
```diff
@@ -0,0 +1,173 @@
...
+	fun rawQuery(key: String, index: Int) = FromRequestDslProperty(template.query(key, index))
+
+	/**
+	 * Unescaped URL path.
+	 * @return dsl property
+	 */
+	fun rawPath() = FromRequestDslProperty(template.path())
+
+	/**
+	 * Unescaped nth value of a URL path (zero indexed) e.g. {{{ request.path.[2]. }}}*
```
OlgaMaciaszek: Add full stop at the end.

specs/spring-cloud-contract-spec-kotlin/src/test/kotlin/org/springframework/cloud/contract/spec/internal/KotlinContractConverterTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+package org.springframework.cloud.contract.spec.internal
+
+import org.junit.jupiter.api.Test
+import org.springframework.cloud.contract.spec.Contract
+import java.io.File
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class KotlinContractConverterTest {
```
OlgaMaciaszek: End class name with `Tests` and not `Test`.

src/checkstyle/checkstyle-suppressions.xml
```diff
@@ -42,7 +42,8 @@
 	<suppress files=".*WiremockServerApplicationTests.*" checks="LineLengthCheck"/>
 	<suppress files=".*ContractVerifierConfigProperties.*" checks="JavadocStyle"/>
 	<suppress files=".*YamlContract.*" checks="JavadocVariable"/>
-	<suppress files=".*HttpHeaders.*" checks=".*"/>
-	<suppress files=".*HttpMethods.*" checks=".*"/>
-	<suppress files=".*HttpStatus.*" checks=".*"/>
+	<suppress files=".*HttpHeaders.*" checks="HideUtility*"/>
```
OlgaMaciaszek: Please remove these suppressions. Instead, fix the javadocs and add private constructors to utility classes.

TYsewyn: If I do that, I'll break backwards compatibility.
See [comment](https://github.com/spring-cloud/spring-cloud-contract/pull/1151#discussion_r313334793) on one of those classes.
I fixed all other checkstyle issues so this the only one that will stay there until the next major release unfortunately.

specs/spring-cloud-contract-spec-kotlin/src/main/kotlin/org/springframework/cloud/contract/spec/internal/CommonDsl.kt
TYsewyn: No, by default classes are final in Kotlin. To be able to extend a class you need to add this `open` modifier.

samples/standalone/kotlin/http-client/src/main/kotlin/com/example/loan/model/LoanApplicationStatus.kt
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.loan.model
+
+enum class LoanApplicationStatus {
+	LOAN_APPLIED,
+	LOAN_APPLICATION_REJECTED,
```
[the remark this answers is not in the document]
TYsewyn: No, it's a trick I learned to avoid multiple line changes while reviewing code.

specs/spring-cloud-contract-spec-kotlin/src/main/kotlin/org/springframework/cloud/contract/spec/internal/FromRequestDsl.kt
marcingrzejszczak: Ah, the checkstyle rules from Boot are not applicable for Kotlin files. That's a shame.

samples/standalone/kotlin/http-client/.gitignore
```diff
@@ -0,0 +1,6 @@
+
+target/
+
+.gradle
```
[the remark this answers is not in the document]
TYsewyn: We're ignoring the `.gradle` directory. The syntax for files is `*.gradle`.

marcingrzejszczak:

samples/standalone/kotlin/http-client/build.gradle.kts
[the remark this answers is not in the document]
sdeleuze: I don't think you can do that for plugins with Gradle Kotlin DSL

Issue #1133, opened 2019-07-08
Support Up-to-date checks (AKA Incremental Build) in Gradle plugin


anatoliy-balakirev:
Enhancement

Gradle plugin doesn't seem to support incremental build => executing `generateContractTests` task every time, which is slowing down execution of completely not relevant unit tests significantly.

Here is some sample project, I created to simulate our setup, where we have a lot of contract tests (we are using Pact jsons as an input):
https://github.com/anatoliy-balakirev/cloud.contract.demo

Those contract tests in the sample project are not really useful, I just wanted to show the case, where there is a project with a lot of them (in real project we have much more, actually).

So if you get into that project and execute following command a couple of times (it's just a unit test):
`./gradlew :test --tests "com.example.demo.service.UserServiceImplTest" --scan`

You'll see that `generateContractTests` is executed every time and takes a couple of seconds even though it's not needed for this particular unit test and its input / output is not changed between runs.
Here is an example of such scan:
https://scans.gradle.com/s/qy3gtncjmopdo/performance/execution
https://scans.gradle.com/s/l2em6bwohe54c/performance/execution

As you can see, there it took almost 2.5 seconds to do something, which is not needed anyway (in our real project it takes up to 10 seconds).

Would be great if plugin could utilise incremental build (check here: https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks) and skip tasks if they are up to date.

marcingrzejszczak:
Great idea @anatoliy-balakirev ! Care for a PR? :)

marcingrzejszczak:
Sorry, I seem too stupid to work with Gradle. What I thought that is working, is not. I will not re-attempt to implement this feature in the upcoming time. If anyone is a Gradle expert, please feel free to try to tackle this problem

shanman190:
@marcingrzejszczak, what's the minimum Gradle version you all are supporting/building against with Hoxton?

marcingrzejszczak:
We're building against 5.5

shanman190:
Also maintaining a 5.0 compatibility like the Spring Boot Gradle plugin?

marcingrzejszczak:
I mean assuming that gradle 5.5 is compatible with gradle 5.0 everything should be fine.

shanman190:
[...]

PR #1165, opened 2019-08-12
Add incremental build support to gradle plugin

anatoliy-balakirev:
In order to make it working - I had to define clear input and output
params for each gradle task.
As part of this task I've also cleaned up some services to define clear
inputs instead of generic `ContractVerifierConfigProperties`.inputs instead of generic `ContractVerifierConfigProperties`.

Fixes gh-1133


pivotal-issuemaster:
@anatoliy-balakirev Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-contract&pullRequestId=1165)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-contract&pullRequestId=1165) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@anatoliy-balakirev Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-contract&pullRequestId=1165)!

shanman190:
So I'll admit that I've had similar issues while working through my refactor of this problem as well. Like you've seen the global mutable state either via the extension updates from within tasks or via the `ext` of the tasks.

This plugin really needs to have a pretty major refactoring to bring it up to what would be considered current standards. You've definitely gone a pretty substantial way there. 

Some things that I'd like to note as pieces we should also consider:

- Plugins are recommended not to automatically create publications unless they are very opinionated about the lifecycle. (Eg: java-gradle-plugin)
- `afterEvaluate` in this plugins case shouldn't really be necessary. The one place I've found that isn't the case is when attaching as a depends on relationship with the `compileTestJava` or `compileTestGroovy` tasks.
- Extension properties can take advantage of being lazy configuration which means that when the property is set by the user later on, then it Cascades through to the tasks that use it as well.
- Could potentially thing forward a little further to build caching. This part I'm working through ideas in relationship to how the `StubsDowloader` downloads the contracts to a temporary directory outside of Gradle's lifecycle and is additionally managing this temp storage with the JVM shutdown cleanup in `TempStorage`. If we let this just be temp storage and copy to an output then it shouldn't be so bad.

These are just some of the things that are coming to mind currently. I'll try to get my branch that I'm working on cleaned up and pushed out for a compare and contrast if anybody is interested?

anatoliy-balakirev:
@shanman190 oh, so you worked on the same in parallel? You sould've mentioend this in the issue, could save us quite some time...

Regarding your points:
1. Interesting, wasn't aware of this (I'm quite new to gradle plugin development; this is actually my first one :) ). But I'm not sure this is related to the incremental build feature, I worked on originally. Maybe worth to create a separate issue / PR for that?
2. and 3. - Nice! Wasn't aware that lazy evaluation can be used with extension properties. Will try to adjust that. I used `afterEvaluate` to actually simulate this laziness in that particular part.
4. Good idea! But again, maybe worth a separate issue / PR?

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
```diff
@@ -62,22 +61,24 @@ public ContractDownloader(StubDownloader stubDownloader,
...
-	 * the inclusion pattern for the downloaded contracts. The JAR with the contracts
-	 * contains all the contracts for all the projects. We're interested only in its
-	 * subset.
-	 * @param config - Plugin configuration that will get updated with the inclusion
-	 * pattern
+	 * Downloads JAR containing all the contracts. The JAR with the contracts contains all
+	 * the contracts for all the projects. We're interested only in its subset.
 	 * @return location of the unpacked downloaded stubs
 	 */
-	public File unpackedDownloadedContracts(ContractVerifierConfigProperties config) {
```
marcingrzejszczak: It's a public method, we should deprecate it instead of removing it.

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
```diff
@@ -62,22 +61,24 @@ public ContractDownloader(StubDownloader stubDownloader,
...
+		}
+		Map.Entry<StubConfiguration, File> unpackedContractStubs = this.stubDownloader
+				.downloadAndUnpackStubJar(this.contractsJarStubConfiguration);
+		if (unpackedContractStubs == null) {
+			throw new IllegalStateException("The contracts failed to be downloaded!");
+		}
+		return unpackedContractStubs.getValue();
 	}
 
-	public ContractVerifierConfigProperties updatePropertiesWithInclusion(
```
marcingrzejszczak: It's a public method, we should deprecate it instead of removing it.

spring-cloud-contract-stub-runner/src/test/groovy/org/springframework/cloud/contract/stubrunner/ContractDownloaderSpec.groovy
```diff
@@ -32,36 +30,31 @@ class ContractDownloaderSpec extends Specification {
 
 	def 'should set inclusion pattern on config when path pattern was explicitly provided with a separator at the beginning'() {
 		given:
-			String contractPath = File.separator + ['a', 'b', 'c', 'd'].join(File.separator)
-			ContractDownloader contractDownloader = new ContractDownloader(stubDownloader,
+			def contractPath = File.separator + ['a', 'b', 'c', 'd'].join(File.separator)
```
marcingrzejszczak: I definitely prefer to have those typed :)

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/ContractsCopyTask.groovy
```diff
@@ -16,18 +16,21 @@
...
-import org.gradle.api.tasks.TaskAction
-import org.gradle.api.tasks.WorkResult
-
-import org.springframework.cloud.contract.verifier.config.ContractVerifierConfigProperties
+import org.gradle.api.Project
+import org.gradle.api.file.CopySpec
+import org.gradle.api.file.Directory
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.*
```
marcingrzejszczak: Let's be explicit about the imports please

marcingrzejszczak:
> So I'll admit that I've had similar issues while working through my refactor of this problem as well. Like you've seen the global mutable state either via the extension updates from within tasks or via the ext of the tasks.

Yup, that's a problem.

> Plugins are recommended not to automatically create publications unless they are very opinionated about the lifecycle. (Eg: java-gradle-plugin)

I guess we're opinionated. We want to have the stub jar publication

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
anatoliy-balakirev: @marcingrzejszczak where is this rule applicable? I mean many classes are used only internally even if methods are public, so if I modify all usages within the project - do I still need to preserve an old method? Is there some module, which is considered public API, which must follow this rule?

spring-cloud-contract-stub-runner/src/test/groovy/org/springframework/cloud/contract/stubrunner/ContractDownloaderSpec.groovy
anatoliy-balakirev: @marcingrzejszczak I saw multiple styles of using this within the project: sometimes it's `def`, sometimes exact type. `def` is quite concise, so I used it here. What is the rule in general? I'm not so familiar with groovy, but I thought that type will actually be derived based on the assigned value (i.e. similar to `var` in the latest java, except for that you can reassign it). That seems to be not so bad... Do you prefer to define concrete types everywhere instead?

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/ContractsCopyTask.groovy
anatoliy-balakirev: Was automatically squashed by IDE. Will fix.

marcingrzejszczak:
> In some other places it's used as an output. I.e. nothing is taken from it, but something is modified in it for further usages (which is not always the case as we are not always sharing the same instance, which is making it even harder to reason about). The fact that this class is mutable, makes this scenario possible everywhere in the code and makes analysing the flow quite hard.

I agree so we can convert the inputs and outputs separately from removing that property configuration.

> Modify as much services as possible to be very explicit about what they need (and there are also some TODOs in the code, suggesting this, which actually encouraged me to proceed);

I dislike being extremely explicit cause afterwards, if there's one more field needed to be used I need to refactor a milion places. Also, often, I have to deprecate an existing public constructor and create a new one. Remember that if we have a public API then we can't just remove stuff. We need to deprecate it and remove it when we have version e.g. 3.0.0

Even if we decide to go that way, I'd prefer that to be done in a separate PR.

> Leave existing services as is (i.e. using that global variable) and only change gradle plugin to dynamically create a new instance of that class every time I need to interact with those services. I was considering this option as well as a bit easier to start with.

Let's go with that if possible. If you think that option 2 makes more sense and won't interfere with the public API then let's go with option 2.

> It is not clear whether service is mutating something inside the ContractVerifierConfigProperties and that should be used as an output;

We can try to find those points and first change the process somehow and only then change the API?

WDYT?

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
marcingrzejszczak: In general if a class is public people might reference it. That's why wherever I can I'm using the package scoped classes. Since this is a library, I have no knowledge of who is actually using this class and how.

> Is there some module, which is considered public API, which must follow this rule?

I'm assuming that all of the public classes follow that rule. At least I try to assume so ;)

spring-cloud-contract-stub-runner/src/test/groovy/org/springframework/cloud/contract/stubrunner/ContractDownloaderSpec.groovy
marcingrzejszczak: Yeah, so I prefer those to be explicit always. I know that they are not done in such a way everywhere, but I try to make those explicit. Groovy is a dynamic language so sometimes the compiler gets lost on what the type is.

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
anatoliy-balakirev: Oh, that's painful... I mean in general, people can use some class / method even if it's not public :) (reflection). So I thought it's usually good to draw line somewhere and say that these classes are part of the public API and are designed to be used externally and those are internal things / implementation details and can be modified. Otherwise refactoring is just impossible.

marcingrzejszczak: > Oh, that's painful... I mean in general, people can use some class / method even if it's not public :) (reflection)

If someone uses reflection then there's not much I can do ;)

>  So I thought it's usually good to draw line somewhere and say that these classes are part of the public API and are designed to be used externally and those are internal things / implementation details and can be modified. Otherwise refactoring is just impossible.

That's a valid point. That's why we try not to have public API as much as possible.

anatoliy-balakirev:
> I dislike being extremely explicit cause afterwards, if there's one more field needed to be used I need to refactor a milion places. 
Yes, there is always some tradeoff... It's either easier to add something without changing many places, but then harder to read / understand what we ended up having, or harder to modify, but easier to reason about each service's API. As at the end developers are spending much more time reading the code than actually writing it - I would still go for the approach with properly defined APIs, TBH.

Anyway, let me try to roll back my `TestGeneratorInput` and changes to public APIs to have smaller PR for now.

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
```diff
@@ -162,19 +189,6 @@ private String wrapWithAntPattern(String path) {
 				+ "**/";
 	}
 
-	private File unpackAndDownloadContracts() {
```
anatoliy-balakirev: Just moved this method up as made it public.

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
```diff
@@ -70,14 +70,43 @@ public ContractDownloader(StubDownloader stubDownloader,
 	 * pattern
 	 * @return location of the unpacked downloaded stubs
 	 */
+	// Use unpackAndDownloadContracts() and createNewInclusionProperties() instead
+	@Deprecated
```
anatoliy-balakirev: These two methods are modifying `ContractVerifierConfigProperties`, so I deprecated them in favour of more straightforward ones (tow separate calls).

spring-cloud-contract-tools/spring-cloud-contract-converters/src/main/groovy/org/springframework/cloud/contract/verifier/converter/RecursiveFilesConverter.groovy
```diff
@@ -43,27 +43,40 @@ import org.springframework.cloud.contract.verifier.wiremock.DslToWireMockClientC
...
 	private final StubGeneratorProvider holder
-	private final ContractVerifierConfigProperties props
 	private final File outMappingsDir
+	private final File contractsDslDir
+	private final List<String> excludedFiles
+	private final String includedContracts
+	private final boolean excludeBuildFolders
 
+	// Use constructor without ContractVerifierConfigProperties
+	@Deprecated
```
anatoliy-balakirev: Decided to keep my changes here to only pass what is actually needed, so deprecating these old constructors.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GenerateServerTestsTask.groovy
```diff
@@ -16,99 +16,113 @@
...
-		File contractsDslDir = contractsDslDir(copyContractsTask, props)
-		if (getConfigProperties().getContractDependency()) {
-			project.logger.debug("Updating the stubs locations for the case where we have a JAR with contracts")
-			props.contractsDslDir = contractsDslDir
-			props.includedContracts = ".*"
-		}
+		logger.info("Generated test sources dir [${generatedTestSourcesDir}]")
+		logger.info("Generated test resources dir [${generatedTestResourcesDir}]")
+		// TODO: Double check this. `ContractDependency` seems to always be `true`. I believe it was wrong here:
+		// if (getConfigProperties().getContractDependency()) {
```
anatoliy-balakirev: Need a feedback on this. Looks like it was a bug here, but I'm not 100% sure...

marcingrzejszczak: Yeah it looks like it. We should check if either group id of a dependency is set or its string notation.

anatoliy-balakirev: @marcingrzejszczak question is: do we really need to do this here? If you check, `GradleContractsDownloader` is updating that `includedContracts` property to some calculated value in case there is group id / we downloaded it. Then this property is only used here. And if in exactly the same scenario we reset this property back to default value, then whole idea of calculating it in the `GradleContractsDownloader` is not clear to me. Moreover, maven plugin doesn't seem to have logic like this.

marcingrzejszczak: Absolutely we don't need to do it here. Actually, we already do it in the `GradleContractsDownloader` as you say, so this shouldn't even be here.

anatoliy-balakirev:
@marcingrzejszczak I've reverted all my changes related to `TestGeneratorInput`. Still, I left my other changes to a couple of services, switching them to more fine grained input. When services are public - left an old signature as well, deprecating it. Can you take a look pls? Is it better now or you would rather do those changes in the separate PR as well (they seem to be quite not big, so decided to leave).

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GenerateServerTestsTask.groovy
anatoliy-balakirev: Ok, removed.

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
```diff
@@ -189,4 +203,26 @@ private String slashSeparatedGroupId() {
 		return this.projectGroupId.replace(".", File.separator);
 	}
 
+	public static class InclusionProperties {
```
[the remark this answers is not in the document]
anatoliy-balakirev: Will add.

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
```diff
@@ -70,14 +70,43 @@ public ContractDownloader(StubDownloader stubDownloader,
...
+	@Deprecated
 	public File unpackedDownloadedContracts(ContractVerifierConfigProperties config) {
 		File contractsDirectory = unpackAndDownloadContracts();
 		updatePropertiesWithInclusion(contractsDirectory, config);
 		return contractsDirectory;
 	}
 
+	// Use createNewInclusionProperties() instead
+	@Deprecated
 	public ContractVerifierConfigProperties updatePropertiesWithInclusion(
```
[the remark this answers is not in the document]
anatoliy-balakirev: Sorry, I'm not sure I understand what you mean. Can you elaborate pls?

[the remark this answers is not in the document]
marcingrzejszczak: You've deprecated the old method so maybe we can start calling the new one from the old one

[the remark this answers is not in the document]
anatoliy-balakirev: Ah, ok. But I'm doing exactly that...

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
```diff
@@ -70,14 +70,43 @@ public ContractDownloader(StubDownloader stubDownloader,
...
 			File contractsDirectory, ContractVerifierConfigProperties config) {
+		final InclusionProperties newInclusionProperties = createNewInclusionProperties(contractsDirectory);
```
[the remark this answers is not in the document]
anatoliy-balakirev: Let's discuss above. I believe this is exactly what I'm doing, but maybe I'm missing something :)

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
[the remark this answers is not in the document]
anatoliy-balakirev: Added

shanman190:
> > Plugins are recommended not to automatically create publications unless they are very opinionated about the lifecycle. (Eg: java-gradle-plugin)

> I guess we're opinionated. We want to have the stub jar publication

So on this one, I've got a story for you. In the case that I have, I'd really honestly go down the path of deprecating this and forcing the end user to specifically add the stubsJar as an artifact that they publish themselves.

Story time...

In our environment (my workplace), we actually have to have every project disable the publications as a result of an oddity we encountered and have just had time to workaround, but not really discuss until now. Basically, our build server is only allowed to push artifacts into our Artifact repository, but not allowed to delete. In several cases, we have teams that are using SCC in the form of CDC as part of the service application Gradle module. This module publishes a Spring Boot jar, source jar, etc. In our case, the stubsJar publication actually conflicts with our normal `maven-publish` publication which had our repository not been configured for write once, then it would have overwritten the root publication. In our case, we just disabled the built-in publishing and added the stubsJar as an artifact to our publication which achieves the same goal without messing up the original pom file.

```
publishing {
    publications {
        maven(MavenPublication) {
            from components.java

            // other artifacts, etc
            artifact sourceJar

            artifact stubsJar
        }
    }
}
```

Now this definitely doesn't need to be a part of this pull request in order to achieve task caching, but I think that it's something to talk about. @marcingrzejszczak, I can write up an issue if you'd prefer to handle it that way or to keep it as a reminder?

anatoliy-balakirev:
Something weird is going on with the build after I synced with original repo...
On the CI (https://circleci.com/gh/spring-cloud/spring-cloud-contract/8633?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)
Build is constantly failing with a lot of errors:
```
[ERROR] Errors: 
[ERROR]   JavaContractConverterSpec.should convert Java DSL with REST to DSL for [#contractFile]:68 NoSuchElement
[ERROR]   JavaContractConverterSpec.should convert Java DSL with REST to DSL for [#contractFile]:68 NoSuchElement
[INFO] 
[ERROR] Tests run: 1093, Failures: 5, Errors: 2, Skipped: 0
[INFO] 
```
While locally it all looks good:
```

@marcingrzejszczak I see those are new tests added by you today. Any idea what's going on there?
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1093, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
```
Moreover, reported failures are in the tests I didn't even touch.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/ContractsCopyTask.groovy
```diff
@@ -16,115 +16,111 @@
...
-			into(outputContractsFolder)
-		}
+		})
 	}
 
-	@CompileStatic
-	private File outputFolder(String root, String suffix) {
-		return getExtension().stubsOutputDir != null ?
-				project.file("${getExtension().stubsOutputDir}/${root}/${suffix}") :
+	static File createTaskOutput(String root, File stubsOutputDir, Project project, String suffix) {
```
shanman190: This appears unused.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GenerateClientStubsFromDslTask.groovy
```diff
@@ -16,66 +16,51 @@
...
-	@CompileDynamic
-	private File contractsDslDir(Task task, ContractVerifierConfigProperties props) {
-		try {
-			return task.ext.contractsDslDir
-		}
-		catch (Exception e) {
-			project.logger.error("Couldn't retrieve the contract dsl property set by the copy contracts task", e)
-			return props.contractsDslDir
-		}
+	static File createTaskOutput(String root, File stubsOutputDir, Project project) {
```
shanman190: This appears unused.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GenerateServerTestsTask.groovy
```diff
@@ -16,99 +16,108 @@
...
  *
  * @since 1.0.0
  */
 @CompileStatic
-class GenerateServerTestsTask extends ConventionTask {
+class GenerateServerTestsTask extends DefaultTask {
+	static final String TASK_NAME = 'generateContractTests'
 
-	File generatedTestSourcesDir
+	@InputDirectory
```
shanman190: If you wanted to make these easier to consume/configure, you can use the `@Nested` qualifer and an object to hold them.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GradleContractsDownloader.groovy
```diff
@@ -26,73 +25,69 @@ class GradleContractsDownloader {
...
-				.withOptions(StubRunnerOptions.fromSystemProps())
-				.withStubRepositoryRoot(extension.contractRepository.repositoryUrl)
-				.withStubsMode(extension.contractsMode)
-				.withUsername(extension.contractRepository.username)
-				.withPassword(extension.contractRepository.password)
-				.withDeleteStubsAfterTest(extension.deleteStubsAfterTest)
-				.withProperties(extension.contractsProperties)
-		if (extension.contractRepository.proxyPort) {
-			options = options.withProxy(extension.contractRepository.proxyHost, extension.contractRepository.proxyPort)
+	static File contractsSubDirIfPresent(File contractsDirectory, Logger logger) {
```
shanman190: I would think this wouldn't need to be static and probably should be private.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/StubRunnerOptionsFactory.groovy
```diff
@@ -0,0 +1,28 @@
+package org.springframework.cloud.contract.verifier.plugin
+
+import groovy.transform.CompileStatic
+import org.springframework.cloud.contract.stubrunner.StubRunnerOptions
+import org.springframework.cloud.contract.stubrunner.StubRunnerOptionsBuilder
+import org.springframework.cloud.contract.stubrunner.spring.StubRunnerProperties
+
+@CompileStatic
+class StubRunnerOptionsFactory {
```
shanman190: Might want to make this `@PackageScoped` that way it doesn't become a part of the public API, huh @marcingrzejszczak?

shanman190:
@anatoliy-balakirev, here's my work in progress if you want to use some of it for inspiration. Mine is going way left field in terms of changes, so yours will make more sense in the short term.

https://github.com/shanman190/spring-cloud-contract/commit/34bf9af5eac2822442cb21c90c4960e9d3188d33

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/ContractsCopyTask.groovy
anatoliy-balakirev: It's used in the `SpringCloudContractVerifierGradlePlugin`

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GenerateClientStubsFromDslTask.groovy
anatoliy-balakirev: The same here. Used in the `SpringCloudContractVerifierGradlePlugin`

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GradleContractsDownloader.groovy
anatoliy-balakirev: Used in the `SpringCloudContractVerifierGradlePlugin`

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GenerateServerTestsTask.groovy
anatoliy-balakirev: Sorry, can you elaborate a bit, pls? How do you think it'll make it easier to configure if I introduce some additional object?

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/StubRunnerOptionsFactory.groovy
anatoliy-balakirev: Done, thx.

anatoliy-balakirev:
@shanman190 Nice, will take a look. Thanks!

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/ContractDownloader.java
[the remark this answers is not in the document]
marcingrzejszczak: Ok I'm blind then ;)

marcingrzejszczak:
>> > Plugins are recommended not to automatically create publications unless they are very opinionated about the lifecycle. (Eg: java-gradle-plugin) ...

@shanman190 hey thanks for the story :)

Let's create a separate issue. Removing that default would be a breaking change :/

marcingrzejszczak:
>Something weird is going on with the build after I synced with original repo... ...

@anatoliy-balakirev eee no idea. Our builds work fine in jenkins. Are your sure you synced with the latest changes?

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GradleContractsDownloader.groovy
shanman190: Ahh, I see now. So something to note about this, is that usually you want the outputs to be a directory within the build folder since:

1) They're gitignored, so changing the data won't result in cascading files into a commit where that doesn't make sense.
2) Further cleans up the input/output relationship to the task. 
2a) You shouldn't need to let the task know about the directory conventions once you're using the lazy configuration API. Essentially, the extension could define the intermediate output directory that contracts should be placed in before being processed by the generation tasks. That output directory would then be the input to the generation tasks.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GenerateServerTestsTask.groovy
shanman190: So kind of along the same lines as what @marcingrzejszczak was talking about by making a class know about a large number of inputs. The same holds true for tasks. While it's most definitely doable, it blurs the api surface. You can kind of see an example of a way to deal with it here:

https://github.com/shanman190/spring-cloud-contract/commit/34bf9af5eac2822442cb21c90c4960e9d3188d33#diff-b998e42b8cb2cccb4ff11ccc92d5f714

anatoliy-balakirev:
@marcingrzejszczak I created another small PR to fix a couple of NPEs and it fails with the same:
https://circleci.com/gh/spring-cloud/spring-cloud-contract/8646?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR is here: https://github.com/spring-cloud/spring-cloud-contract/pull/1170

Both that one and this are based on the latest master, so there is definitely some issue with it / its config for CI:
Screen Shot 2019-08-13 at 10 44 30

Also it started failing when I rebased it yesterday.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GradleContractsDownloader.groovy
anatoliy-balakirev: So, the logic here is:
1. We have an `init` task (used to be part of the `copy contracts`, but I extracted it to the dedicated one to make it incremental). That task is:
1.1 If configured, downloading and unpacking contracts into some temporary directory, which is then used as an input by `copy contracts` task. In this case, that output directory is calculated dynamically by the task itself and then we pick subdirectory from there using this `contractsSubDirIfPresent` method;
1.2 If downloading is not needed - just using default `contracts` directory within `resources` as an input for `copy contracts` task. Again, `contractsSubDirIfPresent` is used there to pick subdirectory if present. That default directory is kind of static and initialised in the `SpringCloudContractVerifierGradlePlugin` as a default output;
2. Copy contracts task is taking that directory from `init` task and copying those contracts to the `build` folder.

So it kind of does what you suggested, but not in the `init` task. Most of the times that `init` task does nothing, actually.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GenerateServerTestsTask.groovy
anatoliy-balakirev: Ok, thanks. Will try to clean it up a bit.

marcingrzejszczak:
Can you rebase against master and try to build it again with the new base image for circleci?

marcingrzejszczak:

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GradleContractsDownloader.groovy
anatoliy-balakirev: I actually moved those properties to the internal nested config class and clean this part up a bit.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GenerateServerTestsTask.groovy
anatoliy-balakirev: Cleaned up.
Also tried to switch `ContractVerifierExtension` to be property based, but got some weird issues there and rolled back. Might be a task for some next PR. Not sure will have more time for this now and it's just a nice to have at the end, not a strict requirement.

anatoliy-balakirev:
@marcingrzejszczak that helped, thanks!

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/GenerateServerTestsTask.groovy
anatoliy-balakirev: Actually, it looks like I'll manage to use properties in the `ContractVerifierExtension` at the end. Already have something working. Only need to fix tests.

anatoliy-balakirev:
@marcingrzejszczak @shanman190 I've updated PR with all the latest changes. Switched extension to use lazy properties as well. Take a look pls, I believe it should be the final version.

anatoliy-balakirev:
The last rebase was to restart CI tests as they seem to be flaky.

marcingrzejszczak:
Congrats @anatoliy-balakirev and @shanman190 on this fantastic PR. Great work!

anatoliy-balakirev:
Thanks! :)

anatoliy-balakirev:
@marcingrzejszczak I just realised that this PR introduced some kind of backward incompatible change in terms of configuring plugin in the `build.gradle.kts` (from here: https://docs.gradle.org/current/userguide/lazy_configuration.html):
```
Note that Gradle Groovy DSL will generate setter methods for each Property-typed property in a task implementation. These setter methods allow you to configure the property using the assignment (=) operator as a convenience.

Kotlin DSL conveniences will be added in a future release.
```

So when using Kotlin DSL, instead of writing this:
```
contracts {
    packageWithBaseClasses = "com.numbrs.referral.controller.contract"
    basePackageForTests = packageWithBaseClasses
    testMode = TestMode.WEBTESTCLIENT
    testFramework = TestFramework.JUNIT5
}
```

We'll have to do this (note `.set()` instead of `=`):
```
contracts {
    packageWithBaseClasses.set( "com.numbrs.referral.controller.contract")
    basePackageForTests.set(packageWithBaseClasses)
    testMode.set(TestMode.WEBTESTCLIENT)
    testFramework.set(TestFramework.JUNIT5)
}
```

Groovy DSL is not affected.
I can create a new PR to add explicit getters if that's an issue.

marcingrzejszczak:
Yeah that's exactly what we had to do here https://github.com/spring-cloud/spring-cloud-contract/commit/3c4772bb6f0dcba6bdfd484657943cb539557e02

It's not a breaking change though since there was no support whatsover for Kotlin. However having consistent behavior would be great. Sure, file another PR :)

anatoliy-balakirev:
I checked this further and looks like currently it's not really possible to do, as Kotlin doesn't support properties with overloaded setters. Here is an original discussion:
https://github.com/gradle/kotlin-dsl/issues/597

And a couple more related links:
https://github.com/gradle/kotlin-dsl/issues/380
https://youtrack.jetbrains.com/issue/KT-24831
https://youtrack.jetbrains.com/issue/KT-4075

So for now Kotlin users will have to write `property.set("value")` instead of `property = "value"`.
So you can adjust it here as well: https://github.com/spring-cloud/spring-cloud-contract/commit/3c4772bb6f0dcba6bdfd484657943cb539557e02

That big assignment (`packageWithBaseClasses = objects.property(String::class).value("com.example.fraud")`) is not needed and can / should be replaced with:
`packageWithBaseClasses.set("com.example.fraud")`

PR #1170, opened 2019-08-13
Fix NPEs

anatoliy-balakirev:
Found a couple of NPEs while testing something else.

### DIFF
--- a/specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/OptionalProperty.java
+++ b/specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/OptionalProperty.java
@@ -44,8 +44,15 @@ public class OptionalProperty implements Serializable, CanBeDynamic {
 	}
 
 	public String value() {
-		return this.value instanceof RegexProperty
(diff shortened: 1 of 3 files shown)


specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/OptionalProperty.java
```diff
@@ -44,8 +44,15 @@ public String optionalPattern() {
 	}
 
 	public String value() {
-		return this.value instanceof RegexProperty
-				? ((RegexProperty) this.value).pattern() : this.value.toString();
+		if (this.value == null) {
+			return "";
```
anatoliy-balakirev: Not sure if this is what needs to be returned here...

marcingrzejszczak: There should be no such option that `null` is passed in. We can leave `""`

marcingrzejszczak:
Can you rebase against master and try to build it again with the new base image for circleci?

anatoliy-balakirev:
@marcingrzejszczak seems to be the same after rebase.

marcingrzejszczak:

PR #1184, opened 2019-08-25
Bugfix/extension dsl

shanman190:
This updates the extension to be completely explict about allowed input types for each method. As such it should also be fully compatible with the kotlin-dsl that is 1.0 GA as part of Gradle 5.0


shanman190:
I'm also working on a test that confirms the Gradle kotlin-dsl support, but it that doesn't make it in with this PR then I'll open another one when I've got it completed.

marcingrzejszczak:
Great, can you ping us once the PR is ready?

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/test/groovy/org/springframework/cloud/contract/verifier/plugin/ContractVerifierIntegrationSpec.groovy
```diff
@@ -54,13 +54,15 @@ abstract class ContractVerifierIntegrationSpec extends Specification {
 		String gradlePluginSysProp = System.getProperty("contract-gradle-plugin-libs-dir")
 		String gradlePluginLibsDir = (gradlePluginSysProp ?: new File("build/").absolutePath.toString()).replace('\\', '\\\\')
 
-		buildFile.write """
-			buildscript {
-				dependencies {
-					classpath fileTree(dir: '$gradlePluginLibsDir', include: '*.jar')
+		initFile.write """
```
shanman190: Switched to using an initscript here as multiple buildscript sections aren't allowed in the Kotlin DSL. They're technically advised against with the Groovy DSL as well.

shanman190:
@marcingrzejszczak, this should now be ready to go. It fixes the bug that you observed where extension methods aren't found on nested extension objects as well as adds a pair of tests for the Kotlin DSL example.

There are some other cleanup things that I would normally do with Gradle, but I feel like I'm stepping too far there. If you're ok with them then I can make the appropriate updates on my side, but there would be a couple of things on your side to do too.

eg:
- Update release.gradle to get off deprecated setters for type Jar
- Consider migration to the [maven-publish](https://docs.gradle.org/current/userguide/publishing_maven.html) plugin
- Continued cleanup of the plugin (I can do another PR for this work as it's unrelated to the effort done here)

marcingrzejszczak:
> Update release.gradle to get off deprecated setters for type Jar

we can remove it cause we don't even perform a release via gradle - we do it via Maven

> Consider migration to the maven-publish plugin

same as the above

> Continued cleanup of the plugin (I can do another PR for this work as it's unrelated to the effort done here)

be my guest :)

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/test/groovy/org/springframework/cloud/contract/verifier/plugin/SampleProjectKotlinSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+package org.springframework.cloud.contract.verifier.plugin
+
+import org.gradle.testkit.runner.BuildResult
+import org.junit.Ignore
+import spock.lang.Stepwise
+
+import static org.gradle.testkit.runner.TaskOutcome.UP_TO_DATE
+
+@Stepwise
+@Ignore
```
marcingrzejszczak: If we ignore it then maybe let's remove it? :)

shanman190: I don't mind removing it all together. I had just tried to make a parallel for each piece of the Java-based variant. Want me to cut it out?

marcingrzejszczak: Yeah let's remove both

shanman190:
>I like migrating more and more stuff to pure Java (hopefully one day we'll migrate everything!).

Yeah, I have the same opinion there. The other part is Gradle actually recommends writing plugins in pure Java anyway, so it's a good transition point. :)

shanman190:
One other thing that I'll note that I observed while working through this. It seems like the checkstyle configuration doesn't quite work right. I kept getting the following when running `./gradlew build`, so eventually I just had to start bypassing the checkstyle tasks. Not sure if there's an underlying issue there or not. I will say that I'm also running on Windows (7 and 10) primarily, though I wouldn't expect OS to be a culprit for a ClassNotFoundException.

```
Execution failed for task ':checkstyleMain'.
> java.lang.ClassNotFoundException: com.puppycrawl.tools.checkstyle.ant.CheckstyleAntTask
```

marcingrzejszczak:
You can remove checkstyle whatsover then

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/test/groovy/org/springframework/cloud/contract/verifier/plugin/SampleProjectKotlinSpec.groovy
shanman190: I've removed the `@Ignored` tests.

shanman190:
Additionally removed:

- release.gradle
- eclipse plugin
- checkstyle configurations

marcingrzejszczak:
Can you rebase against master? The build is broken :|

shanman190:
@marcingrzejszczak, yeah absolutely. Just finished rebased onto latest master.

marcingrzejszczak:
The build still fails on Circle (https://circleci.com/gh/spring-cloud/spring-cloud-contract/9035?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:2.4:install (default-install) on project spring-cloud-contract-gradle-plugin: Failed to install artifact org.springframework.cloud:spring-cloud-contract-gradle-plugin:jar:groovydoc:2.2.0.BUILD-SNAPSHOT: /root/project/spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/build/libs/spring-cloud-contract-gradle-plugin-2.2.0.BUILD-SNAPSHOT-groovydoc.jar (No such file or directory) -> [Help 1]
```

apparently we're not producing a `groovydoc` artifact anymore

shanman190:
Ok, so that appears to have been part of the `release.gradle` file. So what I think was the behavior before, based on looking at the `pom.xml`, is that Maven launches the Gradle `build` task using the maven-exec plugin. This was then running the tests from Gradle's perspective and packaging the classes up (base jar file). As a result/side-effect of using legacy publishing it was also creating the javadoc, groovydoc, and source jars as well which were then scooped up by Maven at the end via the declared artifacts.

A second aspect that I'm curious about now as well is how the [Gradle plugin marker](https://docs.gradle.org/current/userguide/plugins.html?_ga=2.91463322.1077617274.1567442628-1944423831.1567442628#sec:plugin_markers) is being deployed if Maven is the one performing the release? I can see plugin markers for v2.1.1 and v2.1.2, so this appears to have been a recent addition.

I'm thinking that the quick fix is to just restore the `release.gradle` file.

marcingrzejszczak:
>Ok, so that appears to have been part of the release.gradle file. So what I think was the behavior before, based on looking at the pom.xml, is that Maven launches the Gradle build task using the maven-exec plugin. This was then running the tests from Gradle's perspective and packaging the classes up (base jar file). As a result/side-effect of using legacy publishing it was also creating the javadoc, groovydoc, and source jars as well which were then scooped up by Maven at the end via the declared artifacts.

Yeah, we use Maven to publish everything that Gradle had prepared.

> A second aspect that I'm curious about now as well is how the Gradle plugin marker is being deployed if Maven is the one performing the release? I can see plugin markers for v2.1.1 and v2.1.2, so this appears to have been a recent addition.

We're using the `publish-plugin` (`classpath 'com.gradle.publish:plugin-publish-plugin:0.9.10'`). In Maven's pom.xml you can find 

```
<profile>
			<id>central</id>
			<activation>
				<activeByDefault>false</activeByDefault>
			</activation>
			<build>
				<plugins>
					<plugin>
						<groupId>org.codehaus.mojo</groupId>
						<artifactId>exec-maven-plugin</artifactId>
						<version>${exec-maven-plugin.version}</version>
						<executions>
							<execution>
								<id>gradle</id>
								<phase>package</phase>
								<configuration>
									<executable>./gradlew</executable>
									<arguments>
										<argument>clean</argument>
										<argument>${gradle.task}</argument>
										<argument>-PverifierVersion=${project.version}</argument>
										<argument>-Pcentral</argument>
										<argument>publishPlugins</argument>
									</arguments>
								</configuration>
								<goals>
									<goal>exec</goal>
								</goals>
							</execution>
						</executions>
					</plugin>
				</plugins>
			</build>
		</profile>
```

notice the `<argument>publishPlugins</argument>` entry

shanman190:
Ok, now I see what you're talking about. So to be fully complete...

- Maven is the primary build engine
- Maven calls `gradlew build` for integration and release builds, then publishes the resulting artifacts
- Maven calls `gradlew build publishPlugins` during a central release build to publish the Gradle plugin marker.
- Gradle's archive publishing and signing in the release.gradle is not used because Maven does the publishing and signing.

Reasonable summary?

marcingrzejszczak:
Correct!

shanman190:
I've rebased the `release.gradle` script removal out of the history. That should get the build back to green again. 

I didn't want to mess with that part too much since you all already have it working, so IMO it was easier to just revert that change.

shanman190:
Alright, so I see that the tests are failing and I'm pretty sure I know why. However, I'm presently having difficulties with some of the latest snapshots as they have versions that aren't resolving.

```
> Could not resolve all files for configuration ':compileClasspath'.
   > Could not find org.jetbrains.kotlin:kotlin-stdlib-jdk8:${kotlin.version}.
     Required by:
         project : > org.springframework.cloud:spring-cloud-contract-converters:2.2.0.BUILD-SNAPSHOT:20190902.213512-1500 > org.springframework.cloud:spring-cloud-contract-spec:2.2.0.BUILD-SNAPSHOT > org.springframework.cloud:spring-cloud-contract-spec-kotlin:2.2.0
.BUILD-SNAPSHOT
   > Could not find org.jetbrains.kotlin:kotlin-compiler-embeddable:${kotlin.version}.
     Required by:
         project : > org.springframework.cloud:spring-cloud-contract-converters:2.2.0.BUILD-SNAPSHOT:20190902.213512-1500 > org.springframework.cloud:spring-cloud-contract-spec:2.2.0.BUILD-SNAPSHOT > org.springframework.cloud:spring-cloud-contract-spec-kotlin:2.2.0
.BUILD-SNAPSHOT
   > Could not find org.jetbrains.kotlin:kotlin-scripting-compiler-embeddable:${kotlin.version}.
     Required by:
         project : > org.springframework.cloud:spring-cloud-contract-converters:2.2.0.BUILD-SNAPSHOT:20190902.213512-1500 > org.springframework.cloud:spring-cloud-contract-spec:2.2.0.BUILD-SNAPSHOT > org.springframework.cloud:spring-cloud-contract-spec-kotlin:2.2.0
.BUILD-SNAPSHOT
   > Could not find org.jetbrains.kotlin:kotlin-script-util:${kotlin.version}.
     Required by:
         project : > org.springframework.cloud:spring-cloud-contract-converters:2.2.0.BUILD-SNAPSHOT:20190902.213512-1500 > org.springframework.cloud:spring-cloud-contract-spec:2.2.0.BUILD-SNAPSHOT > org.springframework.cloud:spring-cloud-contract-spec-kotlin:2.2.0
.BUILD-SNAPSHOT
```

Same thing was occurring for commons-beanutils through `spring-cloud-contract-converters -> spring-cloud-contract-verifier`.

I'm attempting to force resolve on the versions specified in the bom, but it's setting off a chain reaction as a result...

shanman190:
By using the following and disabling the debug jvm option, I was again successfully able to run both tests that failed. Pending the other issue that I've brought up, which should be recreated with the current build that is in flight, this should be fixed otherwise.

marcingrzejszczak:
Btw we publish the plugin only for ga versions. So snapshots, milestones or rca will not be seen in the plugin portal

shanman190:
I was planning on doing this as part of the next PR, but I've updated the functional test examples to use the new `plugins {}` apply process. I can confirm that the tests are passing on my side. It seemed like the behavior was acting slightly differently based on the OS, so I did some testing within an Ubuntu VM to make sure that things were actually working the way we want them to.

shanman190:
> Btw we publish the plugin only for ga versions. So snapshots, milestones or rca will not be seen in the plugin portal

Sure that makes sense and would help to not confuse users seeing those pieces.

shanman190:
The bootSimple functional sample was used in the documentation. I've restored that and another build is in progress.

shanman190:
@marcingrzejszczak, build pass!

marcingrzejszczak:
Great work, thanks for this contribution!

PR #1223, opened 2019-09-25
SCC Verifier Docker now publishes stubs just once

Dominick1993:
Originally, SCC Verifier Docker published artifacts twice.
Once defined in build.gradle - which also uploaded the original contracts and second time in a publish stubs task defined in gradle plugin. This PR removes the definition from build.gradle and now only the stubs are published and just once.


marcingrzejszczak:
Thanks @Dominick1993 !

PR #1226, opened 2019-09-26
Using same Maven version as Boot

TYsewyn:


marcingrzejszczak:
I wonder why won't boot upgrade to the latest version. Anyways can you add a comment why were doing this?

PR #1311, opened 2020-01-22
Fix StackOverflow exception when stubTrigger runs contract with body based on json file

abialas:


marcingrzejszczak:
Can you please create a test too? And maybe point to the stckoverflow isaue? Thanks!

abialas:
I added the test which verifies there is no more StackOverflow error when trying to extract body value from the json file. There was no issue - I just discovered it when trying to run stubs generated with a body from the json file.

marcingrzejszczak:
So can we close this?

abialas:
If this fix looks ok for you then yes - it can be merged and closed.

marcingrzejszczak:
One of the tests is failing on circle ci, can you please check it out?

PR #1324, opened 2020-02-14
fix for "use the array(String name) form" and "use the object(String …

bojanv55:
fix for "use the array(String name) form" and "use the object(String name) form"


TYsewyn:
Hi @bojanv55! Thank you for your contribution!
Can you add some tests which cover your changes?
I have the feeling this change is unnecessary after PR https://github.com/spring-cloud/spring-cloud-contract/pull/1323.

bojanv55:
You are probably right. I closed it now.

PR #1362, opened 2020-03-30
Add support to resolve credentials from settings.xml

eddumelendez:


marcingrzejszczak:
Hey @eddumelendez ! Thanks for the PR :) Do you think you could write a test for that? You could create a `settings.xml` file somewhere in the test resources of the plugin and assert that we're actually able to decipher that value. What do you think?

eddumelendez:
sure, I will!

eddumelendez:
@marcingrzejszczak I have submit an additional commit with a test not sure if that is enough but TBH it was an interesting one many fixes 😃

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/MavenSettings.java
```diff
@@ -0,0 +1,78 @@
...
+		SettingsDecrypter settingsDecrypter = new DefaultSettingsDecrypter();
+		setField(DefaultSettingsDecrypter.class, "securityDispatcher", settingsDecrypter,
+				new MavenSettings.SpringCloudContractSecDispatcher());
+		return settingsDecrypter;
+	}
+
+	private void setField(Class<?> sourceClass, String fieldName, Object target,
+						  Object value) {
+		try {
+			Field field = sourceClass.getDeclaredField(fieldName);
```
marcingrzejszczak, line 54: Wow, why do we have to do this?

eddumelendez: it is because `DefaultSettingsDecrypter` relies on maven so it will work for the maven plugin but not out of it.

marcingrzejszczak:
Thank you @eddumelendez. As usual, great job! :)

PR #1444, opened 2020-07-11
bug: handle xml with namespaces closes #1427

### LINKED ISSUE #1427: Contract verifier fails to generate tests with xml with namespaces
**Describe the bug**
When defining a contract with an xml that has defined namespaces the generated tests validations are wrongly generated.
I was able to trace the issue to this class in the verifier module ```XmlToXPathsConverter``` but I am not sure about what the solution might be. 
The method ```getNodeValue``` returns an empty string even though the xml has a value.

**Sample**
This repo has an example of the bug.
https://github.com/javiersvg/contract-verifier-xml-bug-demo
It has a minimal contract and runs it with the test runner docker image.

javiersvg:
This PR adds the capabilities to handle xml that contain namespaces.


javiersvg:
Will recreate

PR #1445, opened 2020-07-12
bug: handle xml with namespaces closes #1427

### LINKED ISSUE #1427: Contract verifier fails to generate tests with xml with namespaces

javiersvg:
This PR adds test to validate the expected behavior of the XmlToXPathConverter class when the input is an xml that contains an name space and one that does not contain such namespace.
The implementation of the fix consists in enabling the name space aware property of the DocumentBuilderFactory so that the document instance created from the xml is capable of identify the attributes prefixes and use the DOMNamespaceContect to translate the identified name spaces to the required dom elements.

### DIFF
--- a/spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/util/xml/XmlToXPathsConverter.groovy
+++ b/spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/util/xml/XmlToXPathsConverter.groovy
@@ -16,7 +16,9 @@
 
 package org.springframework.cloud.contract.verifier.util.xml
 
+import com.sun.org.apache.xml.internal.security.utils.DOMNamespaceContext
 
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
Great work @javiersvg , thanks for the PR!

javiersvg:
@marcingrzejszczak I found the same issue is affecting the method used to validate the results. I have 2 possible fixes for it but should I create a new issue or should I just reopen this one?

marcingrzejszczak:
Let's reopen this :)

PR #1455, opened 2020-07-24
Issue 1452

Swierkowski:


marcingrzejszczak:
You need to check out locally the 2.2.x branch and apply the change manualy (e.g. via patch). Otherwise you're trying to merge too many changes as you can see.

Issue #1452, opened 2020-07-23
iso_8601_with_offset predefined pattern


Swierkowski:
It's related to:
https://github.com/spring-cloud/spring-cloud-contract/issues/1296

As pointed in above bug discussion, there should be no limit in the number of digits in the decimal fraction of second. Actually, it is limited to no more than 6.

Currently the regex is:
([0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\.\d{1,6})?(Z|[+-][01]\d:[0-5]\d)

This example does not pass because it has 7 digits:
2020-07-23T11:41:48.9908651Z

Spring Cloud Contract version: 2.2.3

marcingrzejszczak:
Thanks for filing the bug. Are you willing to create a PR that fixes it? :)

PR #1456, opened 2020-07-24
no length limit in ISO8601_WITH_OFFSET second's decimal fraction

Swierkowski:
Fixes gh-1452


marcingrzejszczak:
Congrats on your first PR @Swierkowski ! :tada:

Issue #1470, opened 2020-08-13
Rewrite more Groovy to Java


marcingrzejszczak:
# The rationale

We're looking for help in rewriting all possible production code from Groovy to Java. The Spock tests will remain written in Groovy, however ideally we'd prefer not to have Groovy in any library production code except for the DSL as such.

It's been an ongoing process for years now. The main reason is the compatibility. Gradle comes with a bundled version of groovy and we have a Gradle plugin. Also we depend on groovy version in spring boot. That's all problematic.

It's discouraging for some part of the community to write a fix or a new feature in Groovy rather than in Java.

IDE support for groovy is getting worse every year now. Intellij IDEA introduces new bugs and other IDEs either don't work at all or are even worse.

# What do we want to rewrite?

I'd suggest going through all the modules outside of the `spring-cloud-contract-spec` modules and picking production class after class and rewriting it to java from groovy, running the tests and if the tests work make a commit.

As i said we want to leave the Groovy DSL (groovy-spec module) and the spock tests.

# How to start?

Pick a class / package / module, notify the others in a comment and start converting the classes! Once you're done file a pull request and link it to this issue.

## Checking if code works

Before you file a pull request you can run 

```
$ ./mvnw clean install -Pintegration,docs
```
That way you'll also run the standalone samples that simulate end to end tests. It will also build the documentation and check if the docs aren't broken.

Please note that if you rewrite a class that contains such tags as

```
// tag::foo[]
// end::foo[]
```

that means that most likely that the code within those comments is used in the documentation.

armsargis:
Hi, Groovy DSL should be removed also?

santfirax:
Hello, @marcingrzejszczak I would love to help in this, could it be possible that you provide some guidance for starters?

marcingrzejszczak:
Groovy DSL remains.

I'd suggest going through all the modules outside of the spring cloud contract spec modules and picking production class after class and rewriting it to java from groovy, running the tests and if the tests work make a commit. 

As i said we want to leave the groovy spec and the spock tests.

saket88:
I want to contribute

marcingrzejszczak:
Be my guest. Pick a class or a package and create a pull request where you migrate it to java!

stessy:
I can help as well.

marcingrzejszczak:
Fantastic!

ankurongit:
@marcingrzejszczak  How can I help ?

[...]

PR #1478, opened 2020-08-23
Convert groovy to java (spring-cloud-contract-verifier module)

zhmaeff:
gh-1470


marcingrzejszczak:
Yeah those tests are related to some dependency hell that we have with reactor / spring integration. I'll try to fix it during the week. Thanks for the contribution!

Issue #1167, opened 2019-08-13
Support Gradle Build Cache


shanman190:
Improve build experience with Gradle plugin by supporting Gradle build cache.

https://guides.gradle.org/using-build-cache/

Note: there is also the Gradle build cache for Maven which might be feasible, but I'm not as familiar with Maven and further with this plugin. I do know that with the Gradle plugin, and I'd assume similar is true for Maven, but you can deploy a build cache node (comes with a jar or docker image) somewhere on your network and configure your build engine to use that node as a cache. There are additional capabilities when used with Gradle Enterprise, but again that is not necessary.

shanman190:
All tasks that are able to be cached have been for a bit now. Closing this as completed.

PR #1480, opened 2020-08-23
Convert Gradle plugin from Groovy to Java

shanman190:
References #1470.


marcingrzejszczak:
Looks good so far :)

shanman190:
So the last piece left of this puzzle is the plugin class itself. However, I've got a couple of questions... 

1) As you can see with a couple of the FIXME notes, deleteStubsAfterTest breaks up-to-date checks because the input/output would be cleaned up from outside of the Gradle lifecycle. This would trigger associated tasks to always execute as a result.
2) With the current plugin structure, it is updating the test sourceSet which is also resulting in up-to-date checks to be broken in the default configuration. Which then forces many to use the custom sourceSet support provided via the extension.

What I would propose is to deprecate and completely disable the deleteStubsAfterTest functionality for the Gradle plugin which will fix the first point. To fix the second point, I would propose to create a custom sourceSet (`contracts`) which would get it's own configurations which we could then configure to extend the testImplementation configuration for dependencies and create a custom `Test` task for contracts. Lastly, we would add this `contractsTest` task to the `check` task as a dependsOn which would automatically give IDE support and remove the need to do the kind of odd `idea` plugin modification.

Thoughts?

shanman190:
@marcingrzejszczak, do you have any thoughts on the above topics?

I think I can finish up the plugin class (just have the idea and SourceSet configurations to finish) to make it identical to its current state, but some of what I describe above would make it better in my opinion. So if needed, we can finish up this PR as an exact convert, then try to fix the couple of up-to-date issues from there.

marcingrzejszczak:
Hey @shanman190 . I'm sorry, I must have missed the notification...

> As you can see with a couple of the FIXME notes, deleteStubsAfterTest breaks up-to-date checks because the input/output would be cleaned up from outside of the Gradle lifecycle. This would trigger associated tasks to always execute as a result.

Fair enough. We can even remove it since we're doing breaking changes anyways :shrug: 

> To fix the second point, I would propose to create a custom sourceSet (contracts) which would get it's own configurations which we could then configure to extend the testImplementation configuration for dependencies and create a custom Test task for contracts. 

I like this idea.

> Lastly, we would add this contractsTest task to the check task as a dependsOn which would automatically give IDE support and remove the need to do the kind of odd idea plugin modification.

I like that even more.

> EDIT: Also can somebody elaborate why the extension classes are Serializable? That doesn't make sense to me unless it's related to a static analysis tool.

I remember that one. There was a bug in a version of Gradle and adding that was a workaround. You can delete it.

shanman190:
Probably something along this line will be the end result.

```
apply plugin: "org.springframework.cloud.contract"

contracts {
    /**
      * it would be better to get contracts out of src/test/resources, so that
      * changing a contract doesn't result in unit tests running unecessarily
      **/
    contractsDslDir = file("src/contracts/resources")
}

sourceSets {
    contractTest {
        java.srcDir file("src/contracts/java")
        groovy.srcDir file("src/contracts/groovy")
        resources.srcDirs file("src/contracts/resources")
        compileClasspath += sourceSets.main.output
        runtimeClasspath += sourceSets.main.output
    }
}

configurations {
    contractTestImplementation.extendsFrom testImplementation
    contractTestRuntime.extendsFrom testRuntime
}

task contractTest(type: Test) {
    description = "Runs the contract tests."
    group = "verification"
    testClassesDirs = sourceSets.contractTest.output.classesDirs
    classpath = sourceSets.contractTest.runtimeClasspath
    reports {
        html.destination = file("${buildDir}/reports/contractTests")
        junitXml.destination = file("${buildDir}/contractTest-results")
    }
    mustRunAfter test
}

check.dependsOn(contractTest)
```

marcingrzejszczak:
Oh that's a big change then. Indeed doing it gradually is a good idea. We'd need to change the same for maven (pick both folders) and update the docs.

shanman190:
@marcingrzejszczak, the sourceset part of the equation should be indistinguishable to the user. Changing the contractsDsl directory would definitely be a change and I'd tend to agree that it should be integrated more slowly. The rest we can do today without any end user impact.

As to the Maven side, I also agree. I've honestly never built or updated a Maven plugin, so that would be a new experience. 

For now I'll leave the contracts themselves in their original home, but swap the default sourceSet as described above. Sound good to you?

shanman190:
We might also have to add the `test.output` as an additional item on the classpath as well for transitional purposes. This way we handle the base classes for contract tests.

I'd also be open to adding the new contractDsl directory as a separate PR since it's scope creeped over to the Maven plugin. :)

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/ContractsCopyTask.java
```diff
@@ -0,0 +1,414 @@
...
+
+	private final MapProperty<String, String> contractsProperties;
+
+	private final Property<String> contractsPath;
+
+	// FIXME: I don't think this makes sense given the default configuration... This would only make sense if contractDir == project.projectDir
+	private final Property<Boolean> excludeBuildFolders;
+
+	// FIXME: deleting the stubs after tests breaks incremental builds
+	private final Property<Boolean> deleteStubsAfterTest;
```
shanman190, line 86: So I reviewed this one more thoroughly and found it's usage to be safe from the original Gradle up-to-date checks. The premise is that it's purging the cloned Git repository or Maven dependencies from a temporary folder.

I think it's name could definitely be better or we could simply just configure it to true and not have the parameter. 

Thoughts, @marcingrzejszczak?

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/PublishStubsToScmTask.java
```diff
@@ -0,0 +1,191 @@
...
+class PublishStubsToScmTask extends DefaultTask {
+
+	static final String TASK_NAME = "publishStubsToScm";
+
+	private final Repository contractRepository;
+
+	private final Property<StubRunnerProperties.StubsMode> contractsMode;
+
+	// FIXME: deleting the stubs after tests breaks incremental builds
+	private final Property<Boolean> deleteStubsAfterTest;
```
shanman190, line 63: So I reviewed this one more thoroughly and found it's usage to be safe from the original Gradle up-to-date checks. The premise is that it's purging the cloned Git repository from a temporary folder.

I think it's name could definitely be better or we could simply just configure it to true and not have the parameter.

Thoughts, @marcingrzejszczak?

shanman190:
Ok, at this point the new SourceSet is in place and working quite nicely while pulling the contracts from the `src/test/resources/contracts` directory as before.

Everything is now in Java!

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/ContractsCopyTask.java
marcingrzejszczak: Since we're doing a breaking change release we can rename it and indeed set it by default to `true`. What name would you suggest?

marcingrzejszczak:
> @marcingrzejszczak, the sourceset part of the equation should be indistinguishable to the user. Changing the contractsDsl directory would definitely be a change and I'd tend to agree that it should be integrated more slowly. The rest we can do today without any end user impact.
>As to the Maven side, I also agree. I've honestly never built or updated a Maven plugin, so that would be a new experience.
> For now I'll leave the contracts themselves in their original home, but swap the default sourceSet as described above. This way we can see how that code looks in particular, then we can add in the additional contracts location. Sound good to you?

Sounds great!

> We might also have to add the test.output as an additional item on the classpath as well for transitional purposes. This way we handle the base classes for contract tests.
> I'd also be open to adding the new contractDsl directory as a separate PR since it's scope creeped over to the Maven plugin. :

Sounds good to me.

shanman190:
@marcingrzejszczak, so I think that this is just about ready to go. There are a couple of comments in the review remaining, but this will have removed all of the Groovy code, complete up-to-date/incremental build support, and add build cache support for all tasks that are capable of build cache.

NOTE: `PublishStubsToScm` is not cacheable due to it's own task action.

marcingrzejszczak:
LGTM! I guess we need to update the docs or are they including already changed pieces of code? Also, I can't find any comments in the review, which ones did you refer to?

shanman190:
@marcingrzejszczak, there are two in the `SpringCloudContractVerifierGradlePlugin.java` and one in `ContractsCopyTask.java`.

As to the docs, I'll give them a pass over to make sure everything still reads accurately. As you know most of the changes were internal in nature, but there is one spot that is sourced in via a tag. It'll be good to make sure that that spot still reads accurately to the state of the code here.

shanman190:
Yeah, the docs will need to be updated... I'll try to get the commit together to fixup the docs later this week.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/ContractsCopyTask.java
```diff
@@ -0,0 +1,455 @@
...
+	private final Repository contractRepository;
+
+	private final Property<StubRunnerProperties.StubsMode> contractsMode;
+
+	private final MapProperty<String, String> contractsProperties;
+
+	private final Property<String> contractsPath;
+
+	// FIXME: I don't think this makes sense given the default configuration... This would only make sense if contractDir == project.projectDir
+	private final Property<Boolean> excludeBuildFolders;
```
marcingrzejszczak, line 94: AFAIR this one was related to having a single repository with all the contracts. I'd have to double check what was the exact reason for adding this. I think that when building a JAR with all contracts the build folder would be also automatically attached and put in the stubs jar.

spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/groovy/org/springframework/cloud/contract/verifier/plugin/SpringCloudContractVerifierGradlePlugin.java
```diff
@@ -0,0 +1,389 @@
...
+
+	private void createAndConfigureStubsJarTasks(ContractVerifierExtension extension,
+			TaskProvider<ContractsCopyTask> copyContracts,
+			TaskProvider<GenerateClientStubsFromDslTask> generateClientStubs) {
+		TaskProvider<Jar> task = stubsTask();
+		if (task != null) {
+			// How is this possible? Where can it come from?
+			// TODO: This can only happen if one of the following is true:
+			//   1) plugin was applied twice (allprojects/subprojects/standard apply)
+			//   2) Another task with the same name exists, but was not provided by this plugin
```
marcingrzejszczak, line 252: Yeah that was the case. Someone could have overridden the stubs jar task so we won't override their custom one.

shanman190:
@marcingrzejszczak, I've fixed up the documentation to get it in line with the codebase again. I think all that's left is the one comment in `SpringCloudContractVerifierPlugin.java` centered around the `verifierStubsJar`. I'm still of the opinion that the end user can customize our task, but we should not be using their task.

Last thing, is that if you're ok with all of the changes here, I can rebase on the tip of master and we can get this merged.

docs/src/main/asciidoc/gradle-project.adoc
```diff
@@ -299,25 +319,24 @@ use Spock classes, the class is `spock.lang.Specification`.
...
-`$buildDir/generated-test-resources/contracts`.
+`$buildDir/generated-test-resources/contractTest`.
 * `stubsOutputDir`: Specifies the directory where the generated WireMock stubs from
 the Groovy DSL should be placed.
 * `testFramework`: Specifies the target test framework to be used. Currently, Spock, JUnit 4 (`TestFramework.JUNIT`) and
 JUnit 5 are supported, with JUnit 4 being the default framework.
 * `contractsProperties`: A map that contains properties to be passed to Spring Cloud Contract
 components. Those properties might be used by (for example) built-in or custom Stub Downloaders.
-* `sourceSet`: Source set where the contracts are stored. If not provided will assume `test` (for example, `project.sourceSets.test.java` for JUnit or `project.sourceSets.test.groovy` for Spock).
+* `sourceSet`: Source set where the contracts are stored. If not provided will assume `contractTest` (for example, `project.sourceSets.contractTes.java` for JUnit or `project.sourceSets.contractTest.groovy` for Spock).
```
marcingrzejszczak, line 339: `contractTes` -> `contractTest`

marcingrzejszczak (approved):

marcingrzejszczak:
> @marcingrzejszczak, I've fixed up the documentation to get it in line with the codebase again. I think all that's left is the one comment in SpringCloudContractVerifierPlugin.java centered around the verifierStubsJar. I'm still of the opinion that the end user can customize our task, but we should not be using their task.

Fair enough. If there's an option for the user to customize our stuff we can break that compatibility for 3.0.0. We'll need to provide a proper instruction on how to migrate to the new approach.

> Last thing, is that if you're ok with all of the changes here, I can rebase on the tip of master and we can get this merged.

I'm ok with all the changes.

shanman190:
@marcingrzejszczak, so I came up with an additional point... At the moment the reason for overriding the verifierStubsJar would be to customize it and have it automatically published. Since publishing is now deprecated, it should be reasonable to assume that we can in all cases create our verifierStubsJar as is and if someone wants to publish a different jar using content from various tasks provided by SCC, they can do so. They'd then be free to publish their custom jar instead of the verifierStubsJar without needing to customize our task at all.

I do still agree that we could put a little bit of documentation together to cover some of the customization points that aren't already provided via the extension that end users may have been creating a custom jar to overcome.

Also this will make the enhancement for Gradle Build Cache from #1167.

marcingrzejszczak:
What you wrote makes perfect sense!

shanman190:
@marcingrzejszczak, have you all ever had any trouble with classes in the groovy-stubs directory not being available for compilation? Like the classes are there, but Maven won't bring them in properly during the build?

eg: ./spring-cloud-contract-verifier/target/generated-sources/groovy-stubs/main/org/springframework/cloud/contract/verifier/dsl/wiremock/BaseWireMockStubStrategy.java (+friends)

Environment: 
OS: Windows 10 Enterprise using WSL v1 (Ubuntu 20.04)
Java: AdoptOpenJDK 11.0.8 hotspot [from sdkman]
Maven: using the wrapper provided with the project (3.5.4)

My branch is also up to date with master. Also the reason behind WSL is because I presently don't have a Linux box at hand, but the build to get to this same spot takes around 3 hours on native Windows -- and sometimes fails elsewhere entirely -- and around 6 minutes on Linux.

marcingrzejszczak:
Actually I haven't. I'm not using Windows so I can't even check it :/

However in Circle CI it seems that the build is failing only on the standalone samples

```
[INFO] Build Summary:
[INFO]   Passed: 2, Failed: 2, Errors: 0, Skipped: 0
[INFO] -------------------------------------------------
[ERROR] The following builds failed:
[ERROR] *  dsl/pom.xml
[ERROR] *  webclient/pom.xml
```

You can try running them manually to try to make the build pass.

shanman190:
@marcingrzejszczak, yeah, I was hoping that I could reproduce the failure locally and then go about fixing it with a hopefully feedback loop than spamming everybody with failing circle.ci builds. I'm probably going to end up having to do something along the lines of what you're suggesting since even on the Linux environment I can't get Maven to get past SCC verifier, while the CI setup obviously gets well past that. :shrug:

I'll see how complex it gets to try to force a build down just for the samples. I'm guessing I'm going to need to hack it together given Maven won't be able to give me a verifier jar, so I could build the Gradle plugin. But I'll figure something clever out.

marcingrzejszczak:
>I'll see how complex it gets to try to force a build down just for the samples. I'm guessing I'm going to need to hack it together given Maven won't be able to give me a verifier jar, so I could build the Gradle plugin. But I'll figure something clever out.

Oh that's simple... 

```
# First let's ensure that your gradle changes are there
./mvnw clean install -DskipTests -Pfast -pl :spring-cloud-contract-gradle-plugin

# Install server
pushd samples/standalone/dsl/http-server/
./gradlew clean build publishToMavenLocal
popd

# Install client
pushd samples/standalone/dsl/http-client/
./gradlew clean build publishToMavenLocal
popd
```

shanman190:
@marcingrzejszczak, I'll try that here soon. Since the Gradle pom.xml has a dependency on verifier, I would expect Maven to have tried to build the verifier module main source. Since the main source won't compile due to the generated code linked above -- which is there on the filesystem, but Maven's compiler is stating it's not there --, I was guessing that I'd probably fail before getting to a Gradle plugin install step. This what I meant with clever haha.

shanman190:
🎉 ALL TESTS PASSING! 🎉

shanman190:
@marcingrzejszczak, is there anything else required before we merge this in?

marcingrzejszczak:
All samples in the standalone samples should be using SCC Gradle Plugin.

shanman190:
Ok, that would mean the only functionality used by the `/samples/standalone/we client/http-server` is the `verifierStubsJar`. It was also creating a manual `stubsJar` task with the mappings in the right place from the stubs generated by Rest Docs which is why it _seemed_ like it shouldn't have any need for SCC.

Just to be clear, you still believe that this sample producer should have the SCC Gradle plugin applied?

marcingrzejszczak:
Ah yeah you're right. What i meant is that all of them have a gradle build. I guess we're ready to merge!

marcingrzejszczak:
Fantastic work as usual @shanman190 !

shanman190:
@marcingrzejszczak, I'll double check it after the latest snapshot is built, but a bonus of this is that users _should_ be able to write their base classes in Kotlin with no trouble at all. :)

PR #1512, opened 2020-09-17
Editing and listing fix

Buzzardo:
Edited the new content that came in with the fix for issue 1511.
Also fixed a code listing for JAX-RS. (The tags were misplaced,
and the additional characters broke the listing. I also had to
replace the tab characters with tabs.)


Buzzardo:
Apparently, moving the tags fixes the document but breaks a test.

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-contract/commit/6e951e60fdda57f5e7eec22178415a282390aa4c

PR #1530, opened 2020-10-02
Skip tasks when input directory is empty

shanman190:
When the `contractsDslDir` doesn't exist the tasks would fail. By using `@SkipWhenEmpty` Gradle won't throw an error when the `contractsDslDir` doesn't exist.

### DIFF
--- a/samples/standalone/webclient/http-server/build.gradle
+++ b/samples/standalone/webclient/http-server/build.gradle
@@ -4,6 +4,13 @@ plugins {
 	id "io.spring.dependency-management"
 	id "maven-publish"
 	id "maven"
+	/**
+	 * While the SCC Gradle plugin is not strictly necessary for this project.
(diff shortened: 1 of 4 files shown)


shanman190:
I did add back the plugin for the `/samples/standalone/webclient/http-server` even though it doesn't really need the plugin to perform it's job. However, by having it there it ensures that the edge cases encountered here are also accounted for.

shanman190:
@marcingrzejszczak, fyi, this would presently be a bug with 3.0.0-M4.

shanman190:
@marcingrzejszczak, not quite sure what's going on with the build. It seems to be failing at some place totally unrelated to where the changes are...

shanman190:
Hey, finally the build succeeded. Flaky CI builds... any chance we can get this merged soon, @marcingrzejszczak?

marcingrzejszczak:
Merged! Thanks a lot!

PR #1531, opened 2020-10-04
Migrate DSL contracts to 'contractTest' source set

shanman190:
This continues the effort started in PR #1480 and as discussed there, it was determined that a separate PR should be opened to handle the changed needed for the Maven plugin as well.

### DIFF
--- a/docs/src/main/asciidoc/_project-features-contract.adoc
+++ b/docs/src/main/asciidoc/_project-features-contract.adoc
@@ -103,7 +103,12 @@ contracts {
 
 To get started with writing contracts in Kotlin, you need to start with a (newly created) Kotlin Script file (`.kts`).
 As with the Java DSL, you can put your contracts in any directory of your choice.
-By default, the Maven and Gradle plugins look at the `src/test/resources/contracts` directory.
+By default, the Maven plugin looks at `src/test/resources/contracts` and Gradle plugins look at the
+`src/contractTest/resources/contracts` directory.
+
+NOTE: Since 3.0.0, the Gradle plugin will also look at the legacy
+directory `src/test/resources/contracts` for migration purposes. When contracts are found in this directory, a warning
+will be logged during your build.
 
 You need to explicitly pass the `spring-cloud-contract-spec-kotlin` dependency to your project plugin setup.
 The following example (in both Maven and Gradle) shows how to do so:
--- a/docs/src/main/asciidoc/gradle-project.adoc
+++ b/docs/src/main/asciidoc/gradle-project.adoc
@@ -52,11 +52,11 @@ dependencyManagement {
 }
 
 dependencies {
-	testCompile "org.codehaus.groovy:groovy-all:${groovyVersion}"
+	testImplementation "org.codehaus.groovy:groovy-all:${groovyVersion}"
 	// example with adding Spock core and Spock Spring
-	testCompile "org.spockframework:spock-core:${spockVersion}"
-	testCompile "org.spockframework:spock-spring:${spockVersion}"
-	testCompile 'org.springframework.cloud:spring-cloud-starter-contract-verifier'
+	testImplementation "org.spockframework:spock-core:${spockVersion}"
+	testImplementation "org.spockframework:spock-spring:${spockVersion}"
+	testImplementation 'org.springframework.cloud:spring-cloud-starter-contract-verifier'
 }
 ----
 
@@ -93,11 +93,11 @@ dependencyManagement {
 }
 
 dependencies {
-	testCompile "org.codehaus.groovy:groovy-all:${groovyVersion}"
+	testImplementation "org.codehaus.groovy:groovy-all:${groovyVersion}"
 	// example with adding Spock core and Spock Spring
-	testCompile "org.spockframework:spock-core:${spockVersion}"
-	testCompile "org.spockframework:spock-spring:${spockVersion}"
-	testCompile 'org.springframework.cloud:spring-cloud-starter-contract-verifier'
+	testImplementation "org.spockframework:spock-core:${spockVersion}"
+	testImplementation "org.spockframework:spock-spring:${spockVersion}"
+	testImplementation 'org.springframework.cloud:spring-cloud-starter-contract-verifier'
 }
 ----
 
@@ -118,7 +118,7 @@ buildscript {
 }
 
 apply plugin: 'groovy'
-apply plugin: 'spring-cloud-contract'
+apply plugin: 'org.springframework.cloud.contract'
 
 dependencyManagement {
 	imports {
@@ -127,11 +127,11 @@ dependencyManagement {
 }
 
 dependencies {
-	testCompile "org.codehaus.groovy:groovy-all:${groovyVersion}"
+	testImplementation "org.codehaus.groovy:groovy-all:${groovyVersion}"
 	// example with adding Spock core and Spock Spring
-	testCompile "org.spockframework:spock-core:${spockVersion}"
-	testCompile "org.spockframework:spock-spring:${spockVersion}"
-	testCompile 'org.springframework.cloud:spring-cloud-starter-contract-verifier'
+	testImplementation "org.spockframework:spock-core:${spockVersion}"
+	testImplementation "org.spockframework:spock-spring:${spockVersion}"
+	testImplementation 'org.springframework.cloud:spring-cloud-starter-contract-verifier'
 }
 ----
 ====
@@ -187,7 +187,13 @@ include::{standalone_samples_path}/http-server/build.gradle[tags=repos,indent=0]
 == Add stubs
 
 By default, Spring Cloud Contract Verifier looks for stubs in the
-`src/test/resources/contracts` directory.
+`src/contractTest/resources/contracts` directory. For transitional purposes the plugin
+will also look for contracts in `src/test/resources/contracts`, however, this directory
+is deprecated as of Spring Cloud Contract 3.0.0.
+
+It should also be noted, that with this new Gradle source set, you should also migrate
+any base classes used within your contract tests to `src/contractTest/{language}` where
+`{language}` should be replaced with Java or Groovy as needed for your purposes.
 
 The directory that contains stub definitions is treated as a class name, and each stub
 definition is treated as a single test. Spring Cloud Contract Verifier assumes that it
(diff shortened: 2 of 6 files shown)


shanman190:
This is dependent on PR #1530 to fix some of the tests for the Gradle side. Will revise once that is merged, but we can continue the discussion that started on the original PR.

docs/src/main/asciidoc/_project-features-contract.adoc
```diff
@@ -103,7 +103,7 @@ contracts {
 
 To get started with writing contracts in Kotlin, you need to start with a (newly created) Kotlin Script file (`.kts`).
 As with the Java DSL, you can put your contracts in any directory of your choice.
-By default, the Maven and Gradle plugins look at the `src/test/resources/contracts` directory.
+By default, the Maven and Gradle plugins look at the `src/contractTest/resources/contracts` directory.
```
marcingrzejszczak, line 106: Hmm I don't think this should change for Maven really... It's not really common practice to do it there like that. WDYT?

shanman190: For the Maven side, moving things is kind of a moot point since there is only one test phase generally speaking. If I recall correctly though there was a desire to keep things in line between the two plugins at a high level. 

If you remember, on the Gradle side, if we were to change a contract now as it exists in `src/test/resources` that would trigger the `test` task to rerun because one of its dependent tasks was not up to date.

In this particular scenario, the desire to move them is entirely one sided due to the internal architectures of the build tooling itself. If there aren't any concerns with having two divergent locations for contracts based on the build tooling itself -- with the exception of potentially RestDocs generated contracts -- then I can adjust it back that direction?

marcingrzejszczak: Yeah in Maven we have the incremental compilation in place and it's working fine. I wouldn't change the folder structure.

In Gradle we could look at `src/test/resources/contracts` in order not to destroy the compatibility, however we could suggest that the good place would be `src/contractTest/resources/contracts` - that way you get the proper task up to date resolution. WDYT?

shanman190: Yeah, that sounds like it would be fine. Would the intent be to keep the old path indefinitely or eventually phase it out?

marcingrzejszczak: That's a good question. BTW can you remind me again why `src/test/resources/contracts` is problematic? Which one is the dependent task that is not up to date? Also asking @OlgaMaciaszek for opinion here.

shanman190: Also if helpful, I can throw a quick sample together to show case what I'm meaning. If you think that would help, @marcingrzejszczak?

marcingrzejszczak: Sure, that would be great!

Do I understand it correctly that you're saying that if any resources are changed then it will trigger the contract tests regeneration even though it shouldn't?

shanman190: @marcingrzejszczak, here's that sample that I said I could put together. Hopefully, it'll illustrate the design goal a bit better than just text. 😄 Let me know if it's still puzzling, but all of the steps should be pretty easy to follow from the readme.

https://github.com/shanman190/scc-gh-pull-1531

Note that I'm not using the latest snapshot or milestone in the sample because of PR #1530 being open yet.

shanman190: @marcingrzejszczak, have you been able to get a chance to review the sample?

shanman190: Ok, this has been rebased, so the Gradle plugin should now be fully functioning correctly.

@marcingrzejszczak, knowing that it's your weekend, have you had a chance to review the sample above?

marcingrzejszczak: Hi! Not yet... I'm currently working on some big changes in Spring Cloud Sleuth. Will definitely look into that once I'm done with those :grimacing:

marcingrzejszczak: Ok I'm back! So where were we with this? Did we agree that Maven should not have its path changed and Gradle can? For Gradle we could support both paths (`src/test/resources/contracts` and `src/test/contractTest`) with the annotation / warning that if contracts are found in `src/test/resources/contracts` you're loosing the caching mechanism?

shanman190: Welcome back!

We were thinking that Maven doesn't have a reason to move, correct. With Gradle, the intent was to start out with both available and like you said try to shift over to `src/contractTest/resources/contracts` for better cacheability. 

The last questions that were raised were...
1) From you, why was moving the `src/test/resources/contracts` beneficial for Gradle and how does not doing so impact the cacheability? See sample above for a better walkthrough of how things are impacted taking key notes on the status of the `test` and `contractTest` tasks after each `build` step completed.
2) From me, would we intend to keep Gradle using both directories indefinitely or try to phase out the old location?

Another aspect that I want to put out there for us to consider as well is that if we long term phase out `src/test/resources/contracts` for Gradle, it _may_ make the samples and docs a little more complex to build from a library author standpoint. By no means insurmountable, but just something else that should be considered with all of this.

Hopefully that gives you a good recap on where we're at! :)

marcingrzejszczak: Thanks for the summary. I'll discuss with the broader team this dual support in Gradle. Come to think of it in 3.0.0 we're doing breaking changes, maybe that's a good moment to actually NOT support both folders. What is your opinion on that @shanman190 ?

marcingrzejszczak: Thanks for the feedback! 

> I do think that it would be better to phase out the old location from a Gradle standpoint and only use the new location, but for the point previously mentioned I can see a transitionary period. It would be kind of unfortunate to have to support that form all the way to 4.x if we consider it a breaking change even with a deprecation going into 3.0.

It seems that we might have 4.x sooner than later ;) Let's go with the transitionary period - @OlgaMaciaszek is also on board with this. Deprecation in 3.x - removal in 4.x. We could print out a warning message if we find `src/test/resources/contracts` for Gradle, WDYT?

shanman190: @marcingrzejszczak, haha sounds good!

I'll adjust the Gradle plugin to log a warning when we detect contracts in the old directory and I'll revert my changes with respect to the Maven plugin. How would you recommend handling the docs and samples? We could always just do a small update to recommend contracts be placed in `src/contractTest/resources/contracts` and let the deprecation warning be emitted for the lifetime of 3.x.


A couple of additional bonuses that I can see from all of this effort too are that we can begin putting the base classes in `src/contractTest/*` (due to the inclusion of `sourceSets.test.output` on the `contractTest` classpath they don't have to move right now, though not sure outside of docs how to best recommend moving them there) and it sets us up pretty nicely to add Kotlin test classes in the future. This could start out with SCC generating Kotlin tests from DSL or being able to cross compile. :)

marcingrzejszczak: > I'll adjust the Gradle plugin to log a warning when we detect contracts in the old directory and I'll revert my changes with respect to the Maven plugin. How would you recommend handling the docs and samples? We could always just do a small update to recommend contracts be placed in src/contractTest/resources/contracts and let the deprecation warning be emitted for the lifetime of 3.x.

Very good idea.

> A couple of additional bonuses that I can see from all of this effort too are that we can begin putting the base classes in src/contractTest/* (due to the inclusion of sourceSets.test.output on the contractTest classpath they don't have to move right now, though not sure outside of docs how to best recommend moving them there) and it sets us up pretty nicely to add Kotlin test classes in the future. This could start out with SCC generating Kotlin tests from DSL or being able to cross compile. :)

Maybe updating samples in `spring-cloud-samples/spring-cloud-contract-samples` (branch `3.0.x`)?

shanman190: > Maybe updating samples in spring-cloud-samples/spring-cloud-contract-samples (branch 3.0.x)?

So I just want to clarify, making the associated updates here would require creating a Gradle specific version of the sample. I could go and clean up the Maven styled ones to remove the Gradle files as well. This would create a bit of sample duplication, but I think it's ultimately required as some of the core opinions are changing. WDYT? 

Also if we go this direction, would your thought be to split all of the samples included into their Maven-specific and Gradle-specific counterparts for consistency/accuracy?

marcingrzejszczak: > So I just want to clarify, making the associated updates here would require creating a Gradle specific version of the sample. I could go and clean up the Maven styled ones to remove the Gradle files as well. This would create a bit of sample duplication, but I think it's ultimately required as some of the core opinions are changing. WDYT?

Yeah we would have to refactor those samples. It means A LOT of work really :/ What we could do is to leave the current set up and in one module called `producer_latest_3_0_features` or `consumer_latest_3_0_features` (like this one https://github.com/spring-cloud-samples/spring-cloud-contract-samples/tree/3.0.x/producer_with_latest_2_2_features) we could set up the contracts in proper folders). Then for `4.0.0` we could play around with symbolic links from Maven to Gradle's `contractTests` folders :grimacing: ? Otherwise we would have to have twice as much code as we have already and we do have plenty of code there already. WDYT?

shanman190: I don't have any issue with putting in the work. Haha

Symlinking a bunch of things would sure make me nervous. Haha. Never really sure which tools will work nicely with symlinks and which won't.

For now I'll go ahead and put together a couple of simple samples like you mentioned. Any reason you can see why it couldn't be a Gradle only sample? Since the Maven side isn't changing it could be excluded for now.

marcingrzejszczak: > For now I'll go ahead and put together a couple of simple samples like you mentioned. Any reason you can see why it couldn't be a Gradle only sample? Since the Maven side isn't changing it could be excluded for now.

oh absolutely it could!

shanman190: @marcingrzejszczak, I believe that I've got this updated again with the changes as we discussed. And I've also made a PR using the latest features @ [spring-cloud/spring-cloud-contract-samples#165](https://github.com/spring-cloud-samples/spring-cloud-contract-samples/pull/165).

marcingrzejszczak:
Thank you so much @shanman190 for this PR! As usual fantastic job!

shanman190:
Thanks, @marcingrzejszczak! Sorry that this one took a bit longer than normal. :)

Also noticed a docs typo, so I'll submit a PR to fix that here shortly.

Issue #1542, opened 2020-10-27
RabbitMockConnectionFactoryAutoConfiguration's ConnectionFactory should mock an open Channel


p-ssanders:
**Describe the bug**
Using `spring-cloud-contract-verifier:2.2.3.RELEASE`,
with default configuration (i.e.: `stubrunner.amqp.mockConnection=true`),
when a `ChannelAwareMessageListener` invokes `Channel.basicAck` during a stub test,
then the test fails due to `java.lang.IllegalStateException: Channel closed; cannot ack/nack`.

This happens because the `CachingConnectionFactory.invoke` method in `spring-rabbit:2.2.7.RELEASE` [throws](https://github.com/spring-projects/spring-amqp/blob/3c5aef50cdc5410ab60f7a8bb71ee6b9b071febc/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java#L1175) this exception when the [Channel is closed](https://github.com/spring-projects/spring-amqp/blob/3c5aef50cdc5410ab60f7a8bb71ee6b9b071febc/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/connection/CachingConnectionFactory.java#L1163), _and_ the `RabbitMockConnectionFactoryAutoConfiguration`'s `connectionFactory` bean provider method [doesn't define](https://github.com/spring-cloud/spring-cloud-contract/blob/d1f555f68b541f41f0b099de7ec042162740caaf/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/amqp/RabbitMockConnectionFactoryAutoConfiguration.java#L67) mock behavior for the `Channel.isOpen` method -- so it defaults to `false` (i.e.: always closed).

Defining mock behavior to mock an open channel resolves this issue e.g.:
```
final Channel mockChannel = mock(Channel.class, invocationOnMock -> {
...
    if ("isOpen".equals(invocationOnMock.getMethod().getName())) {
        return true;
    }
...
});
```

marcingrzejszczak:
Thanks for the analysis, are you willing to file a PR to the `2.2.x` branch?

p-ssanders:
Sure would be happy to

p-ssanders:
This issue is resolved by e4ed93acbfb917f6cbafd3d124e7e9a5b978032e

I'll upgrade when that change is released (assuming it will be in 2.2.4.RELEASE).

[...]

PR #1545, opened 2020-10-30
Mock an open channel for stub tests; Fixes gh-1542

p-ssanders:
Adds mock behavior to ensure channel is always open during tests that use `RabbitMockConnectionFactoryAutoConfiguration`'s `ConnectionFactory`


marcingrzejszczak:
Hi @p-ssanders , thanks for the PR. Do you think you could write a test for this?

p-ssanders:
@marcingrzejszczak it occurred to me the changes in this pull request are unnecessary, as the issue the changes intend to resolve (gh-1542) are already resolved by the changes in e4ed93acbfb917f6cbafd3d124e7e9a5b978032e which replace the `CachingConnectionFactory` with a "barebones `ConnectionFactory`"

It was the `CachingConnectionFactory`'s logic that threw the `IllegalStateException` (as documented in gh-1542), but now that this specific implementation was replaced, the invocation of `Channel.basicAck` becomes a no-op as desired.

PR #1547, opened 2020-10-30
Delete only valid mappings before generating stubs at runtime

neptoon:
Improvements over the changes in 6d44c54c. We now only delete files
which contain valid mappings before generating stubs at runtime. This
way we ensure that we actually use the newly generated stubs, but still
are able to keep additional files around to be included via
`body(file(...))`.


spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/HttpServerStubFactory.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.springframework.cloud.contract.stubrunner;
+
+import org.springframework.cloud.contract.stubrunner.provider.wiremock.WireMockHttpServerStub;
+
+/**
+ * Creates {@link HttpServerStub} instances.
+ */
+public class HttpServerStubFactory {
```
marcingrzejszczak, line 24: Why do we need a factory? Why not just leave `new WireMockHttpServerStub()` directly?

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/StubRunnerFactory.java
```diff
@@ -106,6 +111,39 @@ private Path resolvePath(File unpackedLocation) {
 		return path;
 	}
 
+	private void removeCurrentMappings(Path path) {
+		HttpServerStub httpServerStub = HttpServerStubFactory.createServerStub();
```
marcingrzejszczak, line 115: Ok, I think you misunderstood my suggestion. I suggested using the `HttpServerStub` interface like this:

```java
List<HttpServerStub> httpServerStubs = SpringFactoriesLoader.loadFactories(HttpServerStub.class, null);
if (httpServerStubs.isEmpty()) {
  httpServerStubs.add(new WireMockHttpServerStub());
}

...
					if (httpServerStubs.stream().anyMatch(h -> h.isAccepted(file.toFile())) {
						if (log.isDebugEnabled()) {
							log.debug("Deleting file [" + file.toString()
									+ "] since it contains a valid mapping.");
						}

						try {
							Files.delete(file);
						}
						catch (IOException ex) {
							log.warn("Failed to delete file [" + file.toString() + "]",
									ex);
						}
					}

```

something like that... That way we'll reuse whatever the user has provided.

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/provider/wiremock/WireMockHttpServerStub.java
```diff
@@ -237,15 +238,25 @@ private String jsonArrayOfMappings(Collection<String> mappings) {
 
 	@Override
 	public boolean isAccepted(File file) {
-		return file.getName().endsWith(".json");
+		return file.getName().endsWith(".json") && validMapping(file);
```
marcingrzejszczak, line 241: Cool! Can we add a test for that? Or do we have one already?

marcingrzejszczak (changes requested):

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/HttpServerStubFactory.java
neptoon: I was kind of hesitant to add yet another `new WireMockHttpServerStub()` call to the code base. Therefore the idea was to keep the knowledge about the default implementation in one place instead of spreading it over multiple classes, so it could potentially be replaced easier in the future. Might be overkill though :)

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/StubRunnerFactory.java
neptoon: Thanks for the input :)

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/provider/wiremock/WireMockHttpServerStub.java
neptoon: Will add some tests.

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/HttpServerStubFactory.java
neptoon: Removed the factory again in PR update.

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/StubRunnerFactory.java
neptoon: Modified accordingly.

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/provider/wiremock/WireMockHttpServerStub.java
neptoon: Added tests.

spring-cloud-contract-stub-runner/src/test/groovy/org/springframework/cloud/contract/stubrunner/provider/wiremock/WireMockHttpServerStubSpec.groovy
```diff
@@ -76,4 +79,40 @@ class WireMockHttpServerStubSpec extends Specification {
 		cleanup:
 			mappingDescriptor?.stop()
 	}
+
+	def 'should accept a valid mapping'() {
+		given:
+			WireMockHttpServerStub httpServerStub = new WireMockHttpServerStub()
+		when:
```
marcingrzejszczak, line 86: Just FYI you could have written this

```groovy
		when: 
            boolean accepted = httpServerStub.isAccepted(MAPPING_DESCRIPTOR)
		then:
			accepted
```

like this


```groovy
		expect: 
            httpServerStub.isAccepted(MAPPING_DESCRIPTOR)
```

marcingrzejszczak (approved):

marcingrzejszczak:
Thanks for the PR! Let's see how it goes this time ;)

PR #1554, opened 2020-11-09
Fixes for remote contracts

shanman190:
Several fixes for contracts that are fetched remotely via Git.


spring-cloud-contract-tools/spring-cloud-contract-gradle-plugin/src/main/java/org/springframework/cloud/contract/verifier/plugin/ContractsCopyTask.java
```diff
@@ -426,11 +406,16 @@ public Repository(ObjectFactory objects) {
...
 	}
 
-	@Input
+	@Internal
 	MapProperty<String, String> getContractsProperties() {
 		return contractsProperties;
 	}
 
+	@Input
+	Provider<Map<String, String>> getAllContractsProperties() {
```
shanman190, line 415: This might look weird at first, but yes this is intentional to make this the `@Input` and the other `@Internal`. 

The idea is that we desire to have `git.commit` be a part of the map for the purposes of being able to detect up to date checks more precisely when using a remote git repository. If the commit id that the remote is advertising hasn't changed, then it means that there is no work for us to do. While if it has, there is work.

Issue #1423, opened 2020-06-04
Contract to return empty collection generates incorrect producer test / NPE


JosephGaiser:
### Issue
Contracts for returning an empty collection will throw a NPE when generating producer side tests.
When using a workaround to prevent the NPE you will encounter the following issue.
A test will be generated for the contract that does not assert that the collection is empty. Instead it will only assert the status code returned.

### Sample to recreate incorrect test
```
description: Returns an empty collection
request:
  method: GET
  urlPath: /foo
  queryParameters:
    someId: 1234
response:
  status: 200
  bodyFromFile: bodies/EmptyCollectionResponse.json
```
`bodies/EmptyCollectionResponse.json` being  `[]` 
```
@Test
public void validate_foo_1() throws Exception {
	// given:
		WebTestClientRequestSpecification request = given();


	// when:
		WebTestClientResponse response = given().spec(request)
				.queryParam("someId","1234")
				.get("/foo");

	// then:
		assertThat(response.statusCode()).isEqualTo(200);


	// and:
		DocumentContext parsedJson = JsonPath.parse(response.getBody().asString());
}
```
I believe this test should include `assertThatJson(parsedJson).array().isEmpty();` in the `// and` block.

### Sample to reproduce the NPE
```
description: Returns an empty collection
request:
  method: GET
  urlPath: /foo
  queryParameters:
    someId: 1234
response:
  status: 200
  body:
    -
```

exception from `gradle clean generateContractTests`:
```
ask :generateContractTests FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':generateContractTests'.
> java.lang.NullPointerException (no error message)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.3/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 3s
3 actionable tasks: 3 executed
```
`Version 2.2.2.RELEASE`

PR #1556, opened 2020-11-17
Adds handling for empty list and empty map response body

bono007:
Fixes gh-1423

This change allows the response body to be either an empty map (`{}`) or empty collection (`[]`). They can be represented as follows *(yaml and groovy dsl, respectively):*

**empty collection**
```yaml
body: []
```
```groovy
body([])
```
**empty map**
```yaml
body: {}
```
```groovy
body({})
```

---
#### NPE 
The report of the NPE occurs when handling a single null element collection:
```yaml
body:
  - 
```
```groovy
body([null])
```
At first I was going to fix both w/in this code proposal but it became evident that it should it be its own solve - I created #1557 .

### DIFF
--- a/spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/JsonBodyVerificationBuilder.java
+++ b/spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/JsonBodyVerificationBuilder.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.contract.verifier.builder;
 
 import java.math.BigDecimal;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
@@ -102,6 +103,12 @@ class JsonBodyVerificationBuilder implements BodyMethodGeneration, ClassVerifier
 		Object copiedBody = cloneBody(convertedResponseBody);
 		convertedResponseBody = JsonToJsonPathsConverter
 				.removeMatchingJsonPaths(convertedResponseBody, bodyMatchers);
+
+		// If it was a map or list where all elements covered by matchers - json paths
+		// should not include empty check
+		boolean includeEmptyCheck = !mapOrListBodyReducedToEmpty(copiedBody,
+				convertedResponseBody);
+
 		// remove quotes from fromRequest objects before picking json paths
 		TestSideRequestTemplateModel templateModel = hasRequestBody()
 				? TestSideRequestTemplateModel.from(contract.getRequest()) : null;
@@ -109,7 +116,8 @@ class JsonBodyVerificationBuilder implements BodyMethodGeneration, ClassVerifier
 				returnReferencedEntries(templateModel), parsingClosure);
 		JsonPaths jsonPaths = new JsonToJsonPathsConverter(assertJsonSize)
 				.transformToJsonPathWithTestsSideValues(convertedResponseBody,
-						parsingClosure);
+						parsingClosure, includeEmptyCheck);
+
 		DocumentContext finalParsedRequestBody = parsedRequestBody;
 		jsonPaths.forEach(it -> {
 			String method = it.method();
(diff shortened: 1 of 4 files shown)


spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/JsonBodyVerificationBuilder.java
```diff
@@ -102,14 +103,21 @@ Object addJsonResponseBodyCheck(BlockBuilder bb, Object convertedResponseBody,
 		Object copiedBody = cloneBody(convertedResponseBody);
 		convertedResponseBody = JsonToJsonPathsConverter
 				.removeMatchingJsonPaths(convertedResponseBody, bodyMatchers);
+
+		// If it was a map or list where all elements covered by matchers - json paths
+		// should not include empty check
+		boolean includeEmptyCheck = !mapOrListBodyReducedToEmpty(copiedBody,
```
bono007, line 109: This covers the case where the body comes in with fields but they are covered by the matchers so the `convertedResponseBody` is empty. In this case we do not want to add the `isEmpty` check because it is not actually empty, its just "realized" as empty to appease the assertion generation gods.

spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/JsonBodyVerificationBuilder.java
```diff
@@ -123,6 +131,20 @@ Object addJsonResponseBodyCheck(BlockBuilder bb, Object convertedResponseBody,
 		return convertedResponseBody;
 	}
 
+	private boolean mapOrListBodyReducedToEmpty(Object originalBody,
+			Object convertedBody) {
+		int origSize = originalBody instanceof Map ? ((Map) originalBody).size() : -1;
```
bono007, line 136: Well, this addition is, umm how do you say.... gross. Open to suggestions here. What I want to say represent is "is the object a list or map whose size was non-zero before the conversion but is zero after the conversion"

spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/util/JsonToJsonPathsConverter.groovy
```diff
@@ -320,13 +320,15 @@ class JsonToJsonPathsConverter {
 	}
 
 	JsonPaths transformToJsonPathWithTestsSideValues(def json,
-			Closure parsingClosure = MapConverter.JSON_PARSING_CLOSURE) {
-		return transformToJsonPathWithValues(json, SERVER_SIDE, parsingClosure)
+			Closure parsingClosure = MapConverter.JSON_PARSING_CLOSURE,
+			boolean includeEmptyCheck = false) {
```
bono007, line 324: Not super keen on adding a new param, but the default ability of groovy made me ok w/ it. It prevented signature ripple. The other thought was to not introduce the concept of `includeEmptyCheck` and instead have the `JsonBodyVerificationBuilder` remove the paths entry for `$.isEmpty` after calling in here. I landed here as this route was a bit cleaner and its possible we may want to do this same thing in other areas later.

spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/util/JsonToJsonPathsConverter.groovy
```diff
@@ -336,11 +338,11 @@ class JsonToJsonPathsConverter {
 	}
 
 	private JsonPaths transformToJsonPathWithValues(def json, boolean clientSide,
-			Closure parsingClosure = MapConverter.JSON_PARSING_CLOSURE) {
-		if (!json) {
```
bono007, line 340: Before, this would short-circuit when `json` was `null` or an empty collection or map. 

Instead we want to short circuit when `json` is null or an empty collection/map and the  `includeEmptyCheck` is disabled.

spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/JsonBodyVerificationBuilder.java
marcingrzejszczak: I get the intent and the solution looks comprehensible and that's all that matters IMO

spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/builder/JsonBodyVerificationBuilder.java
marcingrzejszczak: Welcome to the world of assertion gods ;)

spring-cloud-contract-verifier/src/main/groovy/org/springframework/cloud/contract/verifier/util/JsonToJsonPathsConverter.groovy
marcingrzejszczak: The Groovy's NullObject is a blessing and a curse at the same time ;)

marcingrzejszczak (approved):

Issue #1204, opened 2019-09-05
@AutoConfigureWireMock does not work on meta-annotations


fdw:
Similar to #1201, `@AutoConfigureWireMock` does not reset the stubbing/mocking/request counters if the test class is annotated with a meta-annotation containing `@AutoConfigureWireMock`.

If you'd rather re-open #1201, please do :)

marcingrzejszczak:
Do you have an example of such usage?

fdw:
For the meta-annotation:
```
@Target(AnnotationTarget.CLASS)
@Retention(AnnotationRetention.RUNTIME)
@Inherited
@SpringBootTest
@AutoConfigureWireMock(port = 0)
@AutoConfigureMockMvc
@TestPropertySource("classpath:application-test.properties")
annotation class IntegrationTest
```

A test that works when directly annotated with `@AutoConfigureWireMock` needs an additional `WireMock.reset()` in the `@Before` method when we use this `@IntegrationTest`.

fdw:
Do you need anything else from me?

marcingrzejszczak:
A sample would be great cause I can't replicate this :grimacing:

fdw:
After testing a lot of different variants, I think my problem is not so much the meta-annotation (that one works as well as the direct annotation), but rather #1027. I'm sorry for the unnecessary work.

fdw:
Unfortunately, this issue still persists.

We have a meta-annotation like this:
```
@Target(AnnotationTarget.CLASS)
@Retention(AnnotationRetention.RUNTIME)
@Inherited
@SpringBootTest
@AutoConfigureWireMock(port = 0)
public @interface ComponentTestAnnotation
```

This meta annotation is used on a base class for tests:

```
@ExtendWith(SpringExtension.class)
@ComponentTestAnnotation
public class BaseComponentTest {
[...]
}
```

Finally, we have an actual test like this:

```
@SpringJUnitWebConfig([...])
public class ActualTest extends BaseComponentTest {
[...]
}
```

And in the appropriate `application.properties` is `wiremock.reset-mappings-after-each-test=true`

However, the mapping and request count are not reset after each test.

With some debugging, the problem seems to lie here:
`if (testContext.getTestClass().getAnnotationsByType(AutoConfigureWireMock.class).length == 0)` (in `WireMockTestExecutionListener`, lines 97-98)
Because the class is not directly annotated with `AutoConfigureWireMock`, the annotation is not found.

[...]

PR #1564, opened 2020-11-20
Support auto configuration on meta-annotations

fdw:
With this change, the stubbing of WireMock will be reset even if only a
meta-annotation is annotated with `@AutoConfigureWireMock`.

Fixes gh-1204


marcingrzejszczak (approved):

marcingrzejszczak:
So the logs look like this

```
org.opentest4j.AssertionFailedError: 

Expecting:
 <"Hello World">
to be equal to:
 <"bar">
but was not.
	at org.springframework.cloud.contract.wiremock.MetaAnnotationWithResetAfterEachTestApplicationTests._03_test(MetaAnnotationWithResetAfterEachTestApplicationTests.java:85)
```

I'm rerunning the build to see if it's repeatable

fdw:
I think I had the same error when I was missing the `DirtiesContext`, if that helps

marcingrzejszczak:
Yeah the merged PR introduced flakeyness in 2.2.x and master. I'll have to look into that before the next release

fdw:
Oh no! I'm so sorry, it was stable on my setup.

Issue #1555, opened 2020-11-13
Wiremock specific application.properties are not working


janis91:
**Describe the bug**
Spring Cloud Version: `Hoxton.SR9`
Maybe I get something wrong, but when I use the `@AutoConfigureWireMock` annotation and have a application.yml in place that looks like the following:
```yml
wiremock:
  server:
    port: 0
    files:
      - "classpath:/mocks"
```

The Wiremock server is not properly configured (the port is not set dynamically and also the stubs are not found). If I set those values directly within the Annotation like `@AutoConfigureWireMock(port = 0, files = ["classpath:/mocks"])` then the configuration works and WireMock has a dynamic port and finds the stubs.

I'd have expected, that the `@PropertyMapping("wiremock.server")` on `@AutoConfigureWireMock` resolves the correct values. But that's apparently not the case.

**Sample**
Two tests are available, one for each described scenarios above.
https://github.com/janis91/wiremock-demo

marcingrzejszczak:
Can you please answer all the required questions described in the issue template? For your convenience let me paste it here

> **Describe the bug**
> Please provide details of the problem, including the version of Spring Cloud that you
are using. 
> 
> **Sample**
> If possible, please provide a test case or sample application that reproduces
the problem. This makes it much easier for us to diagnose the problem and to verify that
we have fixed it.

janis91:
@marcingrzejszczak I have updated the description. Hope the sample helps as well.

bono007:
I see what is going on, the AnnotationsPropertySource is taking precedence over the other profile specific application props (`application-enabled.yml` in the referenced sample). Therefore the `@AutoConfigureWiremock` default attribute values are always being used - unless overridden in the annotation itself. 

![image](https://user-images.githubusercontent.com/28907971/99895761-81ffde00-2c4f-11eb-9d00-f98195c7fd5e.png)

I will continue looking more tomorrow. I want to compare/contrast this behavior with some of these similar `@AutoConfigureXYZ` annotations in Spring Boot.

PR #1566, opened 2020-11-22
Allows @AutoConfigureWireMock to respect user-specified properties 

bono007:
Allows @AutoConfigureWireMock to respect user-specified properties in application.yml

Fixes gh-1555

### DIFF
--- a/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/AutoConfigureWireMock.java
+++ b/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/AutoConfigureWireMock.java
@@ -24,6 +24,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 import org.springframework.boot.test.autoconfigure.properties.PropertyMapping;
+import org.springframework.boot.test.autoconfigure.properties.SkipPropertyMapping;
 import org.springframework.context.annotation.Import;
 
 /**
@@ -40,7 +41,7 @@ import org.springframework.context.annotation.Import;
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
 @Import(WireMockConfiguration.class)
-@PropertyMapping("wiremock.server")
+@PropertyMapping(value = "wiremock.server", skip = SkipPropertyMapping.ON_DEFAULT_VALUE)
 @AutoConfigureHttpClient
 @Inherited
 public @interface AutoConfigureWireMock {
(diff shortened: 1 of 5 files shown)


spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/AutoConfigureWireMock.java
```diff
@@ -40,7 +41,7 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
 @Import(WireMockConfiguration.class)
-@PropertyMapping("wiremock.server")
+@PropertyMapping(value = "wiremock.server", skip = SkipPropertyMapping.ON_DEFAULT_VALUE)
```
bono007, line 44: This is the fix, by comparing/contrasting w/ our tried-true friend [AutoConfigureMockMvc](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/servlet/AutoConfigureMockMvc.java#L63) I was able to see the magic that we needed to apply here.

spring-cloud-contract-wiremock/src/test/java/org/springframework/cloud/contract/wiremock/AutoConfigureWireMockConfigurationInYamlTests.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = WiremockTestsApplication.class,
+		properties = "app.baseUrl=http://localhost:${wiremock.server.port}",
+		webEnvironment = WebEnvironment.NONE)
+@AutoConfigureWireMock
+@ActiveProfiles("customfoo")
+public class AutoConfigureWireMockConfigurationInYamlTests {
```
bono007, line 36: I went ahead and added a test for this case as the other tests did cover the functionality in some form/fashion but they also specified either the `files` or `stubs` to point to content that ultimately returned the same `Hello world` message so we could not really tell if it was the defaults being used or the custom ones. For that reason I also created an explicit resources directory for the `files` and `stubs`.

spring-cloud-contract-wiremock/src/test/resources/application-customfoo.yml
```diff
@@ -0,0 +1,7 @@
+wiremock:
+  server:
+    port: 0
+    stubs:
+      - "file:src/test/resources/custom-from-yml/mappings"
```
bono007, line 5: Again, pointing to these custom dirs assures us that the user overrides here in the yaml are being used.

spring-cloud-contract-wiremock/src/test/resources/custom-from-yml/__files/hello-custom.json
```diff
@@ -0,0 +1,3 @@
+{
+  "message": "Hello Custom One"
```
bono007, line 2: Again, a custom message tailored specific for this case.

marcingrzejszczak:
Great work @bono007, as usual!

Issue #1561, opened 2020-11-20
Dependency convergence error for org.ow2.asm:asm:5.0.4


tillias:
Hello folks,

This is follow up from https://github.com/spring-projects/spring-boot/issues/24218

We're facing dependency convergence error when using spring-boot-starter-test and spring-cloud-starter-contract-verifier.

I have checked json-path:2.4.0 and it was released ages ago, so no coincidence old asm:5.0.4 is used.

Can you please advise what is the best way dealing with it?

Thanks

```
Dependency convergence error for org.ow2.asm:asm:5.0.4 paths to dependency are:
  +-org.springframework.boot:spring-boot-starter-test:2.3.1.RELEASE
    +-com.jayway.jsonpath:json-path:2.4.0
      +-net.minidev:json-smart:2.3
        +-net.minidev:accessors-smart:1.2
          +-org.ow2.asm:asm:5.0.4
and
  +-org.springframework.cloud:spring-cloud-starter-contract-verifier:2.2.3.RELEASE
    +-cglib:cglib:3.2.9
      +-org.ow2.asm:asm:7.0
```

@wilkinsona  mentioned that 

`Generally speaking Spring projects should use the CGLib that's bundled in spring-core rather than pulling it in as a dependency for exactly this sort of reason. Can you please raise this with the Spring Cloud Contract team so that they can explore making that switch?`

bono007:
For a temporary workaround you can exclude the transitive dependency so that the latest version gets used. That is however, currently untested with `json-path -> json-smart -> accessors-smart`. I tested locally on a small project and it worked but that is not a thorough test. I will look at updating the version and getting it tested properly. 

In the meantime the exclusion looks like: 
```
<dependency>
  <groupId>org.springframework.cloud</groupId>
  <artifactId>spring-cloud-starter-contract-verifier</artifactId>
  <scope>test</scope>
  <exclusions>
    <exclusion>
      <groupId>org.ow2.asm</groupId>
      <artifactId>asm</artifactId>
    </exclusion>
  </exclusions>
</dependency>
```

bono007:
For a temporary workaround you can exclude the transitive dependency so that the latest version gets used. That is however, currently untested with `json-path -> json-smart -> accessors-smart`. I tested locally on a small project and it worked but that is not a thorough test. I will look at updating the version and getting it tested properly. 

In the meantime the exclusion looks like: 
```
<dependency>
  <groupId>org.springframework.cloud</groupId>
[...]

PR #1568, opened 2020-11-26
Excludes old asm lib from jsonpath lib

bono007:
Transitively excludes old asm lib from `com.toomuchcoding.jsonassert:jsonassert`(which it pulls in from `jsonpath -> json-smart -> smart-asserter `. 

Fixes gh-1561

This cleans up the `spring-cloud-contract` side of the street. The `spring-boot-starter-test` needs to do the same thing to totally be rid of the old asm lib.

#### Points to note
1. There may be a cleaner, maven-esque way to do this rather than brute force exclusions in the N places that jsonassert is being used. If so, please share.

2. An alternative would have been to exclude the `asm` lib from the `jsonassert` lib and then all we would have to do is bump its version number in here. I chose not to do this as excluding the asm only works under the assumption that an asm lib will be available by the consuming application. We know that is the case here in `spring-cloud-contract` because it is used by Spring applications. We do *not* have that same guarantee w/ `jsonassert`. Removing it from there could break its consumers. This is the less risky move.

### DIFF
--- a/spring-cloud-contract-dependencies/pom.xml
+++ b/spring-cloud-contract-dependencies/pom.xml
@@ -109,6 +109,12 @@
 				<groupId>com.toomuchcoding.jsonassert</groupId>
 				<artifactId>jsonassert</artifactId>
 				<version>${jsonassert.version}</version>
+				<exclusions>
+					<exclusion>
+						<groupId>org.ow2.asm</groupId>
+						<artifactId>asm</artifactId>
+					</exclusion>
+				</exclusions>
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
--- a/spring-cloud-contract-starters/spring-cloud-starter-contract-verifier/pom.xml
+++ b/spring-cloud-contract-starters/spring-cloud-starter-contract-verifier/pom.xml
@@ -41,6 +41,12 @@
 		<dependency>
 			<groupId>com.toomuchcoding.jsonassert</groupId>
 			<artifactId>jsonassert</artifactId>
+			<exclusions>
+				<exclusion>
+					<groupId>org.ow2.asm</groupId>
+					<artifactId>asm</artifactId>
+				</exclusion>
+			</exclusions>
 		</dependency>
 		<dependency>
 			<groupId>cglib</groupId>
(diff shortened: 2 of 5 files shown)


marcingrzejszczak:
Hey @bono007 `com.toomuchcoding.jsonassert:jsonassert` is actually my project so if you feel like we can fix the problem at its core then just file an issue and we can release a new version of it.

bono007:
> Hey @bono007 `com.toomuchcoding.jsonassert:jsonassert` is actually my project so if you feel like we can fix the problem at its core then just file an issue and we can release a new version of it.

@marcingrzejszczak I saw that was your project. I think that is a cleaner fix but my only concern is [non-spring consumers](https://github.com/spring-cloud/spring-cloud-contract/pull/1568#issue-527796766) of it. If that is not an issue I can fix it there. 

BTW, I ❤️ the name `com.toomuchcoding` - too funny.

marcingrzejszczak:
> @marcingrzejszczak I saw that was your project - I heart the name com.toomuchcoding - too funny. I think that is a cleaner fix but my only concern is non-spring consumers of it. If that is not an issue I can fix it there.

We can fix it there and release a major version `1.0.0` that is breaking.

bono007:
@marcingrzejszczak I will submit an MR to `com.toomuchcoding.jsonassert:jsonassert` w/in the hour. Once released I will update this MR accordingly.

bono007:
https://github.com/marcingrzejszczak/jsonassert/pull/35

bono007:
@marcingrzejszczak - my assumption going into my ["fix"](https://github.com/spring-cloud/spring-cloud-contract/commit/de8333deed5f97aca4d8571cbb92941b2e346a45) was that Spring Core brought in the newer version of ASM. However, [further discussion](https://github.com/spring-projects/spring-boot/pull/24264#issuecomment-741797009) on the Spring Boot Starter Test portion of this work has revealed that Spring Core actually shades ASM (like a good library should). 

I think we should remove these exclusions and instead either 

1. Do nothing - the reported issue has a simple workaround
2. Use the shaded jsonassert.

Sorry for the churn on this one but I would rather catch this up front than in the wild if possible.

marcingrzejszczak:
Sure, can you create a PR then that reverts https://github.com/spring-cloud/spring-cloud-contract/commit/de8333deed5f97aca4d8571cbb92941b2e346a45 ? 

As for the shaded jsonassert - that's actually a breaking change as I mentioned before. I'd go with doing nothing and once the boot starter ever changes we'll figure it out

bono007:
Sounds good @marcingrzejszczak . I will not get to it until this evening but I will get it taken care of.

PR #1578, opened 2020-12-10
Removes the exclusions of old asm lib from jsonpath lib.

bono007:
The exclusions were added in gh-1568 in an effort to get rid of the old asm lib that jsonpath brings with it. However, it turns out the issue is bigger than that. Spring Cloud Contract has a direct dependency on `cglib:3.2.9` which in turn brings in a more recent `asm:7.x`. So even before the exclusions, the old `asm` lib was being cancelled out in favor of the new `asm` lib. I originally thought that the more recent asm was being brought in by `spring-core` but it actually shades `cglib` and `asm`. 

While the changes made in gh-1568 were probably harmless - they were not really providing any benefit. Let's remove the changes and revisit later and see if we can remove the direct dependence on cglib at that time as well. 

Reverts gh-1568

### DIFF
--- a/spring-cloud-contract-dependencies/pom.xml
+++ b/spring-cloud-contract-dependencies/pom.xml
@@ -110,12 +110,6 @@
 				<groupId>com.toomuchcoding.jsonassert</groupId>
 				<artifactId>jsonassert</artifactId>
 				<version>${jsonassert.version}</version>
-				<exclusions>
-					<exclusion>
-						<groupId>org.ow2.asm</groupId>
-						<artifactId>asm</artifactId>
-					</exclusion>
-				</exclusions>
 			</dependency>
 			<!-- This has shadowed deps -->
 			<dependency>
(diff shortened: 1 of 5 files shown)


bono007:
@marcingrzejszczak I am not sure how to handle the ticket that the previous pull request closed. I believe are stance on that is now `wont fix` w/ the simple workaround of the consuming application just excluding the `asm` dependency in `spring-boot-starter-test`. 

Someone once told me "dependencies are hard". That is my usual experience w/ them and this roundtrip proved that truth once again. :)

marcingrzejszczak:
> I believe are stance on that is now wont fix w/ the simple workaround of the consuming application just excluding the asm dependency in spring-boot-starter-test.

Yup

bono007:
@marcingrzejszczak I just got a notification that  [json-path](https://github.com/json-path/JsonPath/commit/8ad474399810219f6f2f10433bb977e70f77abfe) finally has a new release. This gives us the ability to affect that project w/ pull requests if needed when we revisit this.

marcingrzejszczak:
WOW, can you file an issue in jsonassert ?

bono007:
@marcingrzejszczak  #1582

PR #1599, opened 2021-01-19
Select value of an `anyOf()` property at random

### LINKED ISSUE #1424: Make anyOf randomly return any of the values, not always the 1st one
**Is your feature request related to a problem? Please describe.**
We're using messaging contract tests quite a bit in our pipelines and we've just found an issue where one of the enum values being passed down had a typo in it. It was never caught by the consumer contract tests because it happened to be the 2nd value declared for the field int he contract, using `anyOf(...)`.

**Describe the solution you'd like**
I'm not sure if it was a deliberate decision to [always use the 1st value provided to `anyOf()`](https://github.com/spring-cloud/spring-cloud-contract/blob/v2.2.3.RELEASE/specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/PatternValueDslProperty.java). I'm wondering if it could instead randomly pick from any of the values declared in the contract. Most of the other DSL properties already leverage random value generation so it seems reasonable to also do it for enums.

Change: (haven't tested at all)
```java
@Override
public T anyOf(String... values) {
    return createAndValidateProperty(RegexPatterns.anyOf(values), values[0]);
}
```
to
```java
@Override
public T anyOf(String... values) {
    final int d = this.random.nextInt(values.length)
    return createAndValidateProperty(RegexPatterns.anyOf(values), values[d]);
}
```

**Describe alternatives you've considered**
Create separate contracts, 1 per enum value. This obviously doesn't scale as we can have many different enums and enum values.

**Additional context**
N/A

sguillope:
The current implementation of `anyOf()` always only selects the first value
when generating a property. It means that other possible enum values are always
ignored when producing a value for the property.

This changes the behaviour so that the value is picked at random from any of
the provided values. It could be considered a breaking changes because of the
change of behaviour although it shouldn't be from the contract perspective.

Fixes #1424

### DIFF
--- a/specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/PatternValueDslProperty.java
+++ b/specs/spring-cloud-contract-spec-java/src/main/java/org/springframework/cloud/contract/spec/internal/PatternValueDslProperty.java
@@ -168,7 +168,7 @@ abstract class PatternValueDslProperty<T extends DslProperty> implements RegexCr
 
 	@Override
 	public T anyOf(String... values) {
-		return createAndValidateProperty(RegexPatterns.anyOf(values), values[0]);
+		return createAndValidateProperty(RegexPatterns.anyOf(values), values[this.random.nextInt(values.length)]);
 	}
 
 }


sguillope:
I couldn't quite see how/where I could add a test for it. Existing tests already use multiple enum values.

marcingrzejszczak:
Thanks @sguillope for the PR!

PR #1607, opened 2021-02-03
Ensure stdout/stderr are logged when using debug mode logging.

shanman190:
Relates to gh-1600.

As you can see in the changes, if `javaexec` were to exit with a return code of 1 an exception was raised and the stdout/stderr that we may have collected during debug logging would have been lost. By usage of the `finally` block we can now ensure that the stdout/stderr from the different forked processes we will create can be captured for debugging information.


shanman190:
@marcingrzejszczak, just to make sure that it's clear this just cleans up the debug output, so that getting to the exception listed in the linked ticket will be easier in the future (eg: the stdout/stderr isn't gobbled up and thrown away). 

Given the real issue with the linked ticket is that the `kotlin-scripting-compiler-embeddable` fell out of sync with the overall project's Kotlin version, there's not really anything that I think we can do from a SC-contract standpoint given the finicky nature of Kotlin itself (as illustrated). 

Just wanting to make sure that it's clear that this isn't "fixing" that issue, but making the discrepancy easier to identify. wdyt?

marcingrzejszczak:
Makes sense @shanman190

Issue #1610, opened 2021-02-09
wiremock.reset-mappings-after-each-test does not work on nested class test


michaldo:
Example: in given form the test pass
When nested class is uncommented, test fails
```
@SpringBootTest(properties = "wiremock.reset-mappings-after-each-test=true")
@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
@AutoConfigureWireMock(port=0)
class WiremockApplicationTests {

	@Value("${wiremock.server.port}")
	String port;

	private WebTestClient webTestClient ;

	@BeforeEach
	void initWebtest() {
		webTestClient = WebTestClient.bindToServer()baseUrl("http://localhost:" + port).build();
	}

//	@Nested
//	class NestedClass {

		@Test
		@Order(1)
		void test200() {
			WireMock.stubFor(WireMock.get("/").willReturn(WireMock.aResponse().withStatus(200)));

			webTestClient.get().exchange().expectStatus().isOk();
		}


		@Test
		@Order(2)
		void test404() {
			webTestClient.get().exchange().expectStatus().isNotFound();
		}
//	}

```
'springCloudVersion', "2020.0.1"

marcingrzejszczak:
I wonder if it's not related to Spring as such. In Contract we're leveraging the Spring's TestExecutionListener. If you run the tests with logs to DEBUG do you see any Contract logs being printed or the WireMockTestExecutionListener is not being called at all?

michaldo:
On nested case I see the following:

> WireMockTestExecutionListener  : No @AutoConfigureWireMock annotation found on [class deleteme.wiremock.WiremockApplicationTests$NestedClass]. Skipping

marcingrzejszczak:
Ah, so we should check for annotations also on the parent class too. We're not supporting nested classes for the time being then.

bono007:
@marcingrzejszczak 

[...]

PR #1611, opened 2021-02-14
Searches enclosing class for @AutoConfigureWiremock

bono007:
Allows @Nested tests to use Wiremock by searching enclosing class for @AutoConfigureWiremock 

Fixes gh-1610

### DIFF
--- a/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockTestExecutionListener.java
+++ b/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockTestExecutionListener.java
@@ -93,7 +93,7 @@ public final class WireMockTestExecutionListener extends AbstractTestExecutionLi
 	}
 
 	private boolean annotationMissing(TestContext testContext) {
-		if (TestContextAnnotationUtils.hasAnnotation(testContext.getTestClass(), AutoConfigureWireMock.class)) {
+		if (!TestContextAnnotationUtils.hasAnnotation(testContext.getTestClass(), AutoConfigureWireMock.class)) {
 			if (log.isDebugEnabled()) {
 				log.debug(
 						"No @AutoConfigureWireMock annotation found on [" + testContext.getTestClass() + "]. Skipping");
--- a/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockTestExecutionListener.java
+++ b/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockTestExecutionListener.java
@@ -16,11 +16,15 @@
 
 package org.springframework.cloud.contract.wiremock;
 
+import java.lang.annotation.Annotation;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.context.ApplicationContext;
-import org.springframework.core.annotation.AnnotationUtils;
+import org.springframework.core.annotation.MergedAnnotation;
+import org.springframework.core.annotation.MergedAnnotations;
+import org.springframework.core.annotation.RepeatableContainers;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.TestContext;
 import org.springframework.test.context.support.AbstractTestExecutionListener;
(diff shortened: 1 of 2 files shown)


spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockTestExecutionListener.java
```diff
@@ -93,7 +97,12 @@ public void afterTestMethod(TestContext testContext) throws Exception {
 	}
 
 	private boolean annotationMissing(TestContext testContext) {
-		if (AnnotationUtils.findAnnotation(testContext.getTestClass(), AutoConfigureWireMock.class) == null) {
+		Annotation annotation = MergedAnnotations
+				.from(testContext.getTestClass(), MergedAnnotations.SearchStrategy.TYPE_HIERARCHY_AND_ENCLOSING_CLASSES,
```
bono007, line 101: We should consider submitting a PR to Spring `AnnotationUtils.findAnnotation` to allow specifying a search strategy. 

There are no other usages of `AnnotationUtils.findAnnotation` in SCC or I would have abstracted this out into SCC AnnotationUtils until it was available in Spring AU.

marcingrzejszczak: cc @sbrannen ?

sbrannen: The proper (and only supported) way to find annotations on a test class, in its type hierarchy, or in its enclosing class hierarchy is to use `org.springframework.test.context.TestContextAnnotationUtils` which provides utilities that transparently honor `@NestedTestConfiguration` semantics.

The `MergedAnnotations.SearchStrategy.TYPE_HIERARCHY_AND_ENCLOSING_CLASSES` strategy is effectively broken and unsupported.

Furthermore, we have no plans to modify `AnnotationUtils` "to allow specifying a search strategy".

bono007: Thx @sbrannen - much cleaner now,

> is to use org.springframework.test.context.TestContextAnnotationUtils which provides utilities that transparently honor @NestedTestConfiguration semantics

Nice! I looked for `TestAnnotationUtils*` in IDEA but did not think to look for `TestContext**` variant.  

New tool in my toolbox:
```
TestContextAnnotationUtils is to AnnotationUtils as ReflectionTestUtils is to ReflectionUtils
```

>The MergedAnnotations.SearchStrategy.TYPE_HIERARCHY_AND_ENCLOSING_CLASSES strategy is effectively broken and unsupported.

Good to know. Are there plans to `@deprecate` the enum value? 

@marcingrzejszczak I have updated accordingly.

sbrannen: > Thx @sbrannen - much cleaner now.

Glad I could help.

> Nice! I looked for `TestAnnotationUtils*` in IDEA but did not think to look for `TestContext**` variant.

Yeah, I had unfortunately already used that name in `org.springframework.test.annotation.TestAnnotationUtils` and didn't want to reuse the name, so I went for `TestContextAnnotationUtils` for the new support.

> ```
> TestContextAnnotationUtils is to AnnotationUtils as ReflectionTestUtils is to ReflectionUtils
> ```

Sure, that's one way to look at it, though `ReflectionTestUtils` is intended to be used by developers in their own tests; whereas, `TestContextAnnotationUtils` is primarily only suited for framework code.

> > The MergedAnnotations.SearchStrategy.TYPE_HIERARCHY_AND_ENCLOSING_CLASSES strategy is effectively broken and unsupported.
> 
> Good to know. Are there plans to `@deprecate` the enum value?

We actually considered doing that for Spring Framework 5.3 but didn't go through with it since some people might actually be using it for special use cases. It's just that the resulting search algorithm does not provide useful semantics within the core Spring Framework itself. That's why I had to introduce `TestContextAnnotationUtils` in order to provide meaningful semantics within `spring-test`.

spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockTestExecutionListener.java
```diff
@@ -97,12 +93,7 @@ public void afterTestMethod(TestContext testContext) throws Exception {
 	}
 
 	private boolean annotationMissing(TestContext testContext) {
-		Annotation annotation = MergedAnnotations
-				.from(testContext.getTestClass(), MergedAnnotations.SearchStrategy.TYPE_HIERARCHY_AND_ENCLOSING_CLASSES,
-						RepeatableContainers.none())
-				.get(AutoConfigureWireMock.class).withNonMergedAttributes().synthesize(MergedAnnotation::isPresent)
-				.orElse(null);
-		if (annotation == null) {
+		if (TestContextAnnotationUtils.findMergedAnnotation(testContext.getTestClass(), AutoConfigureWireMock.class) == null) {
```
sbrannen, line 96: As of Spring Framework 5.3.3, you should be able to replace that with `TestContextAnnotationUtils.hasAnnotation(testContext.getTestClass(), AutoConfigureWireMock.class)`.

spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockTestExecutionListener.java
bono007: Thx again for the background info and details.

marcingrzejszczak:
Thanks for the contribution @bono007 !

PR #1618, opened 2021-03-01
Fix Clone on Windows instructions

plamentotev:
The following command, suggested by the README

```
git clone -c core.longPaths true https://github.com/spring-cloud/spring-cloud-contract.git
```

fails because Git recognizes `true` as the repo to be cloned.

Add equal sign (`=`) to fix this issue. This way Git will properly recognize `true` as the value for `core.longPaths`


marcingrzejszczak:
Thanks @plamentotev for your contribution!

Issue #1587, opened 2020-12-21
3.0.0.RC1 generated test compares kafka header to String, though it is byte array


Stmated:
I am using version 3.0.0-RC1 of Spring Cloud, and I'm trying to create a test for Kafka. In it comes a header test.
So the contract says something like:

```
outputMessage:
  sentTo: "booked"
  headers:
    BOOK-HEADER: "foo"
```

But when you create a Kafka `ProducerRecord` it only takes headers as byte-array values, not string. So when I send a message from somewhere in the code like:
```
var producerRecord = new ProducerRecord<String, BookingResult>("booked", ...);
producerRecord.headers().add("BOOK-HEADER", "foo".getBytes(StandardCharsets.UTF_8));
this.kafkaTemplate.send(producerRecord);
```

And the generated test is:
```
ContractVerifierMessage response = contractVerifierMessaging.receive("booked", ...);
assertThat(response.getHeader("BOOK-HEADER").toString()).isEqualTo("foo");
```

It will fail with something along the line of "Expected 'foo' given '[B@58372a00'".

Something I could do is create a custom `ContractVerifierMessaging` bean that does this header conversion for me, but feels like it should be able to be solved in a better way?

I can understand if this is not so simple, since the contract does state a string, and that is not what is returned.
But I can't really see how this is supposed to be done properly?

bono007:
It is an interesting problem because header values in Kafka are `byte[]` (as you have pointed out) and the content in those `byte[]` is application specific. There is no type information associated with them. The reasoning why that is the case can be seen in [KIP-82](https://cwiki.apache.org/confluence/display/KAFKA/KIP-82+-+Add+Record+Headers#KIP82AddRecordHeaders-ProducerRecord%3CK,H,V%3E,ConsumerRecord%3CK,H,V%3E).

Additionally, the code that builds that [particular clause](https://github.com/spring-cloud/spring-cloud-contract/blob/master/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/builder/MessagingHeadersThen.java#L59) always calls `toString()` on the header value (except when number) which will always result as the memory address of the `[B@58372a00'"` (the header value `byte[].toString()`) being compared against the contract value (`"foo"`). 

It does feel like the header value must be converted in the incoming message in order to match the generated test rules. Again, because the headers can be any arbitrary object and are application specific, converting them all to String would not work in all cases. 

[...]

PR #1623, opened 2021-03-12
Add support for comparing byte[] Kafka message header values

bono007:
Fixes gh-1587

| ℹ️ I took the simple route of assuming UTF-8 and doing the conversion if its a byte[]. This fixes the issue at hand. |
| --- |


marcingrzejszczak:
great work as usual @bono007 !

Issue #854, opened 2019-01-18
Dynamic properties in `queryParameters` section are not parsed


tsayen:
It looks like dynamic properties don't work for `request.urlPath.queryParameters` when defining contract in Groovy

Relevant part of the contract looks like that:

```
           request {
                method 'GET'
                urlPath('/advise') {
                    queryParameters {
                        parameter('referenceDate': $(
                                consumer(regex(iso8601WithOffset())),
                                producer(execute("todayAtNoon()"))
                        ))
                    }
                }
            }
```
And `todayAtNoon()` is a method in the base test class.
This results in generated of test code:
```
           // when:
			Response response = given().spec(request)
					.queryParam("referenceDate","todayAtNoon()")
```
This of course causes 400 Bad Request, as we're literally sending `"todayAtNoon()"` string to the producer.

The version is `2.0.2.RELEASE`

OlgaMaciaszek:
Hello @tsayen . Thanks for reporting it. It's a good idea. Marking as enhancement and adding to backlog.

tsayen:
Hi @OlgaMaciaszek, I don't really see this issue as an enhancement, it's rather a bug. 

The documentation clearly states you can do something like that: https://cloud.spring.io/spring-cloud-static/spring-cloud-contract/1.1.2.RELEASE/#_executing_custom_methods_on_server_side

But it doesn't work

tsayen:
There is also issue #709, which is about dynamic properties, but in request headers. In that case `execute()` works correctly, it just doesn't work for `queryParameters`

PsHype:
I also have this issue , I think it is bug

markgargan:
Spent half a day looking at this before realising that it's distinct to the queryParameter and not my syntax of the 
producer(execute("myBaseClassTestMethod()"))

It's been a year any update on it?

marcingrzejszczak:
As you can see nothing has changed. We're open to accepting PRs though.

markgargan:
If you can give me a few pointers I'd take a look?

marcingrzejszczak:
[...]

PR #1646, opened 2021-04-25
Parse ExecutionProperty in queryParameters

mariuszdb:
Fixes gh-854. Query param isn't quoted if it's an ExecutionProperty, similarly to https://github.com/spring-cloud/spring-cloud-contract/blob/c323e3e105810498a2bb9869d87eb144d5c8d0d5/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/builder/JaxRsRequestHeadersWhen.java#L46-L67


marcingrzejszczak:
Great job @mariuszdb !

PR #1663, opened 2021-05-22
Fix request headers occurring after absent header not being added to generated test method

### LINKED ISSUE #1647: All headers missing from generated contract test when using absent() for one of them
I'm using Spring Cloud Contract 2.2.6.RELEASE.

I've written a contract that uses `absent()` for one of the headers:

```groovy
Contract.make {
    request {
        method 'POST'
        urlPath '/foobar'
        headers {
            header(authorization(), absent())
            contentType(MediaType.APPLICATION_JSON_VALUE)
        }
        body([...])
    }
    response {
        status FORBIDDEN()
    }
}
```

I'm expecting the generated test to contain the `Content-Type` header, but there are no headers at all:
```java
RequestSpecification request = given()
		.body(...);
```

When I switch the order of the headers, then the generated test does include the `Content-Type` header:

```groovy
Contract.make {
    request {
        method 'POST'
        urlPath '/foobar'
        headers {
            contentType(MediaType.APPLICATION_JSON_VALUE)
            header(authorization(), absent())
        }
        body([...])
    }
    response {
        status FORBIDDEN()
    }
}
```

```java
RequestSpecification request = given()
		.header("Content-Type", "application/json")
		.body(...);
```

Based on my investigation, this is caused by the `return` statement in this bit of code from `org.springframework.cloud.contract.verifier.builder.MockMvcHeadersGiven`:
```java
private void processInput(BlockBuilder bb, Headers headers) {
	Iterator<Header> iterator = headers.getEntries().iterator();
	while (iterator.hasNext()) {
		Header header = iterator.next();
		if (ofAbsentType(header)) {
			return;
		}
		if (iterator.hasNext()) {
			bb.addLine(string(header));
		}
		else {
			bb.addIndented(string(header));
		}
	}
}
```

cselagea:
Fixes #1647

### DIFF
--- a/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/builder/MockMvcHeadersGiven.java
+++ b/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/builder/MockMvcHeadersGiven.java
@@ -44,7 +44,7 @@ class MockMvcHeadersGiven implements Given {
 		while (iterator.hasNext()) {
 			Header header = iterator.next();
 			if (ofAbsentType(header)) {
-				return;
+				continue;
 			}
 			if (iterator.hasNext()) {
 				bb.addLine(string(header));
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
Thank you so much for your contribution!

Issue #1655, opened 2021-05-04
Enable CORS in stub-runner with stubrunner configurations


szatyinadam:
In our project we use stub-runner to stub backend based on our contracts and the front-end team can develop independently using these stubs.
We use the springcloud/spring-cloud-contract-stub-runner Docker image to serve these stubs but we can not enable CORS globally in the stub runner.
Wiremock has a config for this (`.stubCorsEnabled(true)`), maybe stub-runner can support this as well.

Our preferred solution is a stubrunner config which can get from `${ADDITIONAL_OPTS}`, for example: `"--stubrunner.enableCors=true"`

TheCoffeeOfficial:
This could be a standalone enhancement or the part of this https://github.com/spring-cloud/spring-cloud-contract/issues/1268

marcingrzejszczak:
I like the idea behind #1268. Let's go with such an approach. Feel free to file a pr and I'll review it

PR #1687, opened 2021-08-04
Add stub CORS enabled option to WireMock start

szatyinadam:
Configure WireMock's CORS configuration via environmental variable.

#1655


szatyinadam:
@marcingrzejszczak What is the preferred way, class to implement config options? What config options required from the [WireMockConfiguration](https://github.com/tomakehurst/wiremock/blob/master/src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java) ? I assume some of them is not needed, for example `threadPoolFactoy`.

marcingrzejszczak:
We already have the `wiremock` properties -> https://github.com/spring-cloud/spring-cloud-contract/blob/c42293b14f6636a459205e7594e328f0aba6ee79/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockConfiguration.java#L327-L452 . We could make this bean public and reuse it and ensure that all options available here http://wiremock.org/docs/running-standalone/ can be set

marcingrzejszczak:
@szatyinadam how is it going with the PR? :)

marcingrzejszczak:
I'm closing the PR due to inactivity

aosorio1:
@marcingrzejszczak  I was waiting eagerly that this PR got merged just to see that it was closed. Right now, it's not possible to disable WireMock request journal via properties and this makes it difficult to use WireMock in scenarios where journal should be disabled such as load testing. Any chances of reopening this issue?

PR #1688, opened 2021-08-06
Add support for v4 UUIDs, and allow UUIDs to be case insensitive

jamietanna:
As some web services expect to receive a valid UUIDv4, rather than a rough UUID
pattern, we should provide a convenience method to generate them.

We should add test coverage to make it clear that this is a strict UUIDv4.

Also, allow uppercase UUID generation, as it's perfectly valid for a UUID to be
formatted with uppercase or lowercase, as per RFC4122 Section 3.


marcingrzejszczak (approved):

jamietanna:
Wow, thanks for the speedy merge! :raised_hands:

PR #1693, opened 2021-08-10
fixes #1639: run with unprivileged user and bump java version to 8.0.292.hs

### LINKED ISSUE #1639: Docker image - root required
**Describe the bug**
The Docker image runs on root by default which is both insecure and prevents the container from running in secured environments (like Gitlab CI).
If the user is specified (i.e. by providing `--user 1000` flag), the container fails.

**Sample**
```
❯ docker container run --user 1000 --rm -ti springcloud/spring-cloud-contract:3.0.2 ./gradlew clean build

ERROR: JAVA_HOME is set to an invalid directory: /root/.sdkman/candidates/java/current/

Please set the JAVA_HOME variable in your environment to match the
location of your Java installation.
```

bastiandoetsch:
Running with unprivileged user (scc, UID 1000, /home/scc) now. Also bumped java minor version to 8.0.292.


marcingrzejszczak:
Hey @bastiandoetsch can we do a PR against 3.0.x branch?

bastiandoetsch:
Sure. Gimme a minute.

bastiandoetsch:
I'll open a new one, as it's gonna be better to branch from 3.0.x and then apply my changes.

Issue #1630, opened 2021-03-19
Body value with a Map<String, Set<String>> creates a wrong assertion


martinvisser:
**Describe the bug**
Using the following contract an assertion is created which won't work. Changing `Set.of` to `List.of` will make the assertion in the created test work however.

**Sample**
The contract, using the Java DSL:
```java
import java.util.Collection;
import java.util.List;
import java.util.Map;
import java.util.Set;
import java.util.function.Supplier;

import org.springframework.cloud.contract.spec.Contract;

@SuppressWarnings("unused")
class map_response implements Supplier<Collection<Contract>> {
    @Override
    public Collection<Contract> get() {
        return List.of(Contract.make(c -> {
            c.description("get map");
            c.name("assert map response");
            c.request(r -> {
                r.urlPath("/some-path");
                r.method(r.GET());
            });
            c.response(r -> {
                r.status(r.OK());
                r.headers(h -> h.contentType(h.applicationJson()));
                r.body(r.value(Map.of("key", Set.of("value1", "value2")))); // With List.of("value1", "value2") a correct assertion is generated
            });
        }));
    }
}
```
Generated assertion, which fails:
```java
DocumentContext parsedJson = JsonPath.parse(response.getBody().asString());
assertThatJson(parsedJson).field("['key']").isEqualTo("[value1, value2]");
```

However, if we change `Set.of("value1", "value2")` to `List.of("value1", "value2")` the following assertion is created, which works:
```java
DocumentContext parsedJson = JsonPath.parse(response.getBody().asString());
assertThatJson(parsedJson).array("['key']").arrayField().isEqualTo("value1").value();
assertThatJson(parsedJson).array("['key']").arrayField().isEqualTo("value2").value();
```

In both cases the json would be identical:
```json
{
  "key": [
    "value1",
    "value"
  ]
}
```

A `List` is split into multiple assertions, but a `Set` isn't? I would expect the same behaviour actually.

marcingrzejszczak:
Yeah it should work with sets. Please use the list until this gets fixed.

PR #1710, opened 2021-08-31
#1630 Fix assertion for Map<String, Set<String>>

dehasi:
Relates to [#1630](https://github.com/spring-cloud/spring-cloud-contract/issues/1630)

I added a test that reproduces the issue and added a fix that achieves the expected behavior.

@marcingrzejszczak in the ticket you say `it should work with sets`, 
but a set is an unordered collection, but arrays in JSON are ordered.

Do we check the order of elements in JSON?  If not, just accept my PR :)

If yes, I'm hesitating that we can treat sets as lists. Maybe we have to check `contains` for sets, but not `equals`.
I can work on it, but I need this to be clarified.

Also
`SingleTestGeneratorSpec` fails in the `main` branch as well, so it's not related to my changes.

### DIFF
--- a/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/util/MapConverter.java
+++ b/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/util/MapConverter.java
@@ -16,9 +16,11 @@
 
 package org.springframework.cloud.contract.verifier.util;
 
+import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -41,6 +43,7 @@ import org.springframework.cloud.contract.verifier.template.TemplateProcessor;
  *
  * @author Marcin Grzejszczak
  * @author Stessy Delcroix
+ * @author Ravil Galeyev
  * @since 1.1.0
  */
 public class MapConverter {
@@ -123,8 +126,8 @@ public class MapConverter {
 		else if (value instanceof Map) {
 			return convert((Map) value, function, parsingFunction);
 		}
-		else if (value instanceof List) {
-			return ((List) value).stream().map((v) -> transformValues(v, function, parsingFunction))
+		else if (value instanceof List || value instanceof Set) {
+			return ((Collection) value).stream().map((v) -> transformValues(v, function, parsingFunction))
 					.collect(Collectors.toList());
 		}
 		return transformValue(function, value, parsingFunction);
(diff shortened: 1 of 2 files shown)


spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/util/MapConverter.java
```diff
@@ -123,8 +126,8 @@ else if (parsed instanceof List) {
 		else if (value instanceof Map) {
 			return convert((Map) value, function, parsingFunction);
 		}
-		else if (value instanceof List) {
-			return ((List) value).stream().map((v) -> transformValues(v, function, parsingFunction))
+		else if (value instanceof List || value instanceof Set) {
```
marcingrzejszczak, line 129: Can we actually make this `value instanceof Collection` ?

spring-cloud-contract-verifier/src/test/groovy/org/springframework/cloud/contract/verifier/builder/SpringTestMethodBodyBuildersSpec.groovy
```diff
@@ -1934,6 +1939,41 @@ World.'''"""
...
+		Contract contractDsl = Contract.make {
+			description('get map')
+			name('assert map response')
+			request {
+				method 'GET'
+				urlPath '/some-path'
+			}
+			response {
+				status OK()
+				body(Map.of("key", Set.of("value1", "value2")))
```
marcingrzejszczak, line 1954: Change this to `body(["key": ["value1", "value2"] as Set])`

marcingrzejszczak (changes requested):
In general LGTM with small suggestions of changes, although I think we should apply this fix against the 3.0.x branch.

spring-cloud-contract-verifier/src/test/groovy/org/springframework/cloud/contract/verifier/builder/SpringTestMethodBodyBuildersSpec.groovy
dehasi: Nice shot! And a good improvement of my `groovy` skills. 
Fixed, thank you!

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/util/MapConverter.java
dehasi: I thought about it.
If we treat all collections (including stack or queue) as `Map<Key, List<>>` we definitely can. But should we? 
Json has arrays and arrays has an order, that's why I'm not sure that it's a good idea 
to treat `Map<Key,Set<>>` as `Map<Key,List<>>`. 
On the other hand, from the "business logic" POV we know that nobody would use `Map<Key, Queue<>>` or some other collections.

So, @marcingrzejszczak  It's up to you. I updated, but can revert if you wish :)

marcingrzejszczak (approved):

marcingrzejszczak:
Congrats @dehasi on your first contribution to Spring Cloud Contract!

PR #1713, opened 2021-09-02
fix #1711: Possible NPE in JMS stubs

### LINKED ISSUE #1711: Possible NPE in JMS stub verifier
**Describe the bug**

It is possible to pass null _headers_ into `JmsStubMessages.send()` method. 

    @Override
	public void send(Object payload, Map headers, String destination, YamlContract contract) {
		jmsTemplate.send(destination, session -> {
			Message message = createMessage(session, payload);
			setHeaders(message, headers);
			return message;
		});
	}

In this case there will be an error, because it tries to get entrySet():

    private void setHeaders(Message message, Map<String, Object> headers) {
		for (Map.Entry<String, Object> entry : headers.entrySet()) {
			String key = entry.getKey();
			Object value = entry.getValue();
			try {
				if (value instanceof String) {
					message.setStringProperty(key, (String) value);
				}
				else if (value instanceof Boolean) {
					message.setBooleanProperty(key, (Boolean) value);
				}
				else {
					message.setObjectProperty(key, value);
				}
			}
			catch (JMSException ex) {
				throw new IllegalStateException(ex);
			}
		}
	}

Similar bug: #1705

mygordienko:
@pivotal-cla This is an Obvious Fix

### DIFF
--- a/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/util/MapConverter.java
+++ b/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/util/MapConverter.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.contract.verifier.util;
 
+import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


marcingrzejszczak (approved):

marcingrzejszczak:
Sth is wrong with the commit history. There should be one commit

marcingrzejszczak:
@mygordienko can you rebase against 3.0.x branch please?

Issue #1727, opened 2021-11-08
Metadata duplication on stub mappings generation


gabbigum:
**Describe the bug**

Hi, all. I have the following issue. When creating a contract with the wiremock webhook extension the generated stub mappings duplicate the **postServeActions**.

This contract:
```
request:
  method: POST
  url: /provision
  body:
    "machine": "aws-VM"
    "callbackURL": "http://localhost:8080/callback"
  headers:
    Content-Type: application/json
response:
  status: 200
metadata:
  wiremock:
    stubMapping:
      postServeActions:
        webhook:
          method: PATCH
          url: "http://localhost:8080/callback"
          headers:
            Content-Type: "application/json-patch+json"
          body: "{ \"host\" : \"AWS\" }"
```

Generates this stub mapping: 
```
{
  "id" : "3466c327-f926-4ecb-bad2-ae328516e0f4",
  "request" : {
    "urlPath" : "/provision",
    "method" : "POST",
    "headers" : {
      "Content-Type" : {
        "equalTo" : "application/json"
      }
    },
    "bodyPatterns" : [ {
      "matchesJsonPath" : "$[?(@.['machine'] == 'aws-VM')]"
    }, {
      "matchesJsonPath" : "$[?(@.['callbackURL'] == 'http://localhost:8080/callback')]"
    } ]
  },
  "response" : {
    "status" : 200
  },
  "uuid" : "3466c327-f926-4ecb-bad2-ae328516e0f4",
  "postServeActions" : [ {
    "name" : "webhook",
    "parameters" : {
      "method" : "PATCH",
      "url" : "http://localhost:8080/callback",
      "headers" : {
        "Content-Type" : "application/json-patch+json"
      },
      "body" : "{ \"host\" : \"AWS\" }"
    }
  }, {
    "name" : "webhook",
    "parameters" : {
      "method" : "PATCH",
      "url" : "http://localhost:8080/callback",
      "headers" : {
        "Content-Type" : "application/json-patch+json"
      },
      "body" : "{ \"host\" : \"AWS\" }"
    }
  } ]
}

```

Am I doing something wrong in my stubMapping definition or this is a valid bug? If it is a bug I would like to get some info which files to check/debug and help and fixing it.

Thanks in advance!

marcingrzejszczak:
[...]

PR #1728, opened 2021-11-09
Fixes gh-1727 metadata duplication on stub mappings generation

gabbigum:


spring-cloud-contract-tools/spring-cloud-contract-converters/src/main/java/org/springframework/cloud/contract/verifier/wiremock/DefaultWireMockStubPostProcessor.java
```diff
@@ -43,7 +43,10 @@ public StubMapping postProcess(StubMapping stubMapping, Contract contract) {
 		StubMapping stubMappingFromMetadata = stubMappingFromMetadata(wireMockMetaData.getStubMapping());
 		stubMapping.setResponse(mergedResponse(stubMapping, stubMappingFromMetadata));
 		if (stubMappingFromMetadata.getPostServeActions() != null) {
-			setPostServeActions(stubMapping, stubMappingFromMetadata);
+			//FIXES ISSUE-1727
```
marcingrzejszczak, line 46: You can remove this line cause the commit message will tell us which issue got fixed by this.

spring-cloud-contract-tools/spring-cloud-contract-converters/src/test/groovy/org/springframework/cloud/contract/verifier/wiremock/DslToWireMockClientConverterSpec.groovy
```diff
@@ -1074,6 +1074,57 @@ class DslToWireMockClientConverterSpec extends Specification {
...
 	}
 
+
+	@Issue("1727")
+	def "should create WireMock JSON() with no duplicate metadata fields"() {
+		given:
+			def converter = new DslToWireMockClientConverter()
+		and:
+			File file = tmpFolder.newFile("stub_mapping_duplicate.groovy")
+			file.write("org.springframework.cloud.contract.spec.Contract.make {\n" +
```
marcingrzejszczak, line 1084: IN Groovy you can do a multline string like this

```groovy
"""
{ 
"foo" : "bar"

}
"""
```

spring-cloud-contract-tools/spring-cloud-contract-converters/src/test/groovy/org/springframework/cloud/contract/verifier/wiremock/DslToWireMockClientConverterSpec.groovy
```diff
@@ -1074,6 +1074,57 @@ class DslToWireMockClientConverterSpec extends Specification {
 			response.statusCodeValue == 204
 	}
 
+
+	@Issue("1727")
```
marcingrzejszczak, line 1078: :+1:

spring-cloud-contract-tools/spring-cloud-contract-converters/src/main/java/org/springframework/cloud/contract/verifier/wiremock/DefaultWireMockStubPostProcessor.java
```diff
@@ -43,7 +43,10 @@ public StubMapping postProcess(StubMapping stubMapping, Contract contract) {
...
+			if(stubMapping.getPostServeActions() == null) {
```
marcingrzejszczak, line 47: So what should happen if we already have post serve actions? Shouldn't we merge the two?

gabbigum: From my observations the _postServeActions_ do not belong to _stubMapping_ by default, rather we are **injecting** them to the _stubMapping_ after we extract them from the contract at _postProcess_ stage. 

I see that’s a valid point about **merging** them, since the _postProcess_ gets invoked twice in this scenario. But _setPostServeActions()_ actually **appends** the _postServeActions_ to the existing _postServeActions_ of _stubMapping_ and does not merge them like _setMetadata_ does, since the underlying collection is **list** not a **map** which causes the issue.

 And from what I've seen the extracted _postServeActions_ from the _stubMappingFromMetadata_ are the same every time. Please correct me if I'm wrong. That's my first contribution to an open source project and I highly appreciate your feedback. :)

marcingrzejszczak: > I see that’s a valid point about merging them, since the postProcess gets invoked twice in this scenario. But setPostServeActions() actually appends the postServeActions to the existing postServeActions of stubMapping and does not merge them like setMetadata does, since the underlying collection is list not a map which causes the issue.

why are we calling this twice?

> And from what I've seen the extracted postServeActions from the stubMappingFromMetadata are the same every time. Please correct me if I'm wrong. That's my first contribution to an open source project and I highly appreciate your feedback. :)

Ok, I'm starting to remember this... Indeed if they are already set they shouldn't be set again. The question remains why are we calling post process twice. You're doing great work with the contribution :)

gabbigum: Great question. I've traced the execution flow very carefully and examined this:

1. convertASingleContract(1) calls postProcessStubMapping(2)
2. postProcessStubMapping(2) calls super.postProcessStubMapping(3)
3. super.postProcessStubMapping(3) calls the overriden defaultStubMappingPostProcessing(4) that's in the DslToWireMockClientConverter since processors are empty()
5. The overriden defaultStubMappingPostProcessing(4) calls postProcess(5) - which appends the postServeActions to the stub mapping - 1st time
6. After which we call defaultStubMappingPostProcessing is called once again at postProcessStubMapping - which repeats step 5 - thus appending postServeActions to stubMapping - 2nd time

The issue lies in the
```	@Override
	public StubMapping postProcessStubMapping(StubMapping stubMapping, Contract contract) {
		return super.postProcessStubMapping(stubMapping, contract); // that makes sure only 1 call is invoked in the post process stage
		// apply the default WireMock processor as the last one
		//return defaultStubMappingPostProcessing(mapping, contract); // removing this 
	}
```

I've commented out the last line that was calling the defaultStubMappingPostProcessing for the second time since it has been already executed in the super.postProcessStubMapping and removed the `if(stubMapping.getPostServeActions() == null)` that I've previously written.

Is there any reason you've called the super.postProcessStubMapping and then again defaultStubMappingPostProcessing in the first place? Is there anything that I'm breaking? I've ran all the tests and everything was passing. Thanks for the quick feedback :)

marcingrzejszczak: > Is there any reason you've called the super.postProcessStubMapping and then again defaultStubMappingPostProcessing in the first place? Is there anything that I'm breaking? I've ran all the tests and everything was passing. Thanks for the quick feedback :)

Honestly? I don't remember ;) If the tests are passing that we could assume that things are working fine

gabbigum: I've applied the changes we discussed. Will wait for your feedback once again. Thank you very much!

marcingrzejszczak (approved):

spring-cloud-contract-tools/spring-cloud-contract-converters/src/main/java/org/springframework/cloud/contract/verifier/wiremock/DefaultWireMockStubPostProcessor.java
gabbigum: It seems that my IDE deleted new line in one of the files and that's why build fails with checkstyle error. I'll make another commit.

marcingrzejszczak:
Congrats @gabbigum on your contribution :)

gabbigum:
Thank you very much!

Issue #1734, opened 2021-11-16
publishStubsToScm fails with NoClassDefFoundError


p-ssanders:
**Describe the bug**
With 
- Spring Cloud Contract 2020.0.4
- Spring Boot 2.5.6
- spring-cloud-contract-gradle-plugin 3.0.4
- Gradle 7.3

`publishStubsToScm` gradle task fails when targeting a remote git repository with the error:
```
Caused by: java.lang.IllegalStateException: Exception occurred while cloning repo
	at org.springframework.cloud.contract.stubrunner.GitRepo.cloneProject(GitRepo.java:126)
...
Caused by: java.lang.NoClassDefFoundError: com/jcraft/jsch/agentproxy/connector/SSHAgentConnector
	at org.springframework.cloud.contract.stubrunner.GitRepo$JGitFactory$1.createDefaultJSch(GitRepo.java:339)
```

**Sample**
A project to reproduce this problem can be found at https://github.com/p-ssanders/scc-publishstubstoscm-example
Run `./gradlew clean build publishStubsToScm --stacktrace` to reproduce

PR #1735, opened 2021-11-21
Enable SSH Agent support.

shanman190:
Fixes gh-1734.


shanman190:
@marcingrzejszczak, I've reviewed the code itself and I agree that this is the right fix to enable SSH Agent support. However, something to note is that JNAUSocketFactory.java tries to open the native `libc` library to create sockets. `libc` is unavailable on Windows machines, so effectively SSH Agent support only works for Linux or Macintosh machines.

I'm experimenting with some alternatives to enable Windows support in addition to the existing setup.

shanman190:
Tried a sample project using junixsocket (Java 7+) and native Unix sockets (JDK 17; [JEP-380](https://openjdk.java.net/jeps/380)) to no success. Tried on both Windows 10 Home and Windows 10 Enterprise with similar errors. `ssh-agent` works in both cases including compatibility with `ssh-add` and `ssh -T git@github.com` . Both command line use cases work without any issues, but since Windows lacks a `libc` equivalent and the above two alternatives don't work either, I think SSH Agent support with JGit is going to be limited to Linux and Macintosh only.

Issue #1740, opened 2021-12-03
Add given/providerStates part when generate a RequestResponsePact from a Contract


mathieu-amblard:
I am using the `ToFileContractsTransformer` in version 3.0.4 to generate a PACT file from a Contract for HTTP as described in the reference documentation : https://docs.spring.io/spring-cloud-contract/docs/3.0.0/reference/htmlsingle/#how-to-generate-pact-from-scc.

My Contract looks like this :
```groovy
package com.test

import org.springframework.cloud.contract.spec.Contract

Contract.make {
    description("""
Some description...
""")
    request {
       ...
    }
    response {
        ...
    }
}
```

The generate PACT file looks like this :
```json
{
    "provider": {
        "name": "..."
    },
    "consumer": {
        "name": "..."
    },
    "interactions": [
        {
            "description": "Some description",
            "request": {
                ...
            },
            "response": {
                ...
            }
        }
    ],
    "metadata": {
        "pactSpecification": {
            "version": "3.0.0"
        },
        "pact-jvm": {
            "version": "4.0.10"
        }
    }
}
```

Firstly, I would like to know if there is any reason to not have a `providerState` as it is done for the `MessagePact` here : https://github.com/spring-cloud/spring-cloud-contract/blob/90172436be6590732c2ec70a1e98ce6716fd3c11/spring-cloud-contract-tools/spring-cloud-contract-pact/src/main/java/org/springframework/cloud/contract/verifier/spec/pact/MessagePactCreator.java#L58 or if it is an oversight.

Secondly, if this is an oversight, is it possible to add it the same way, and with maybe an input field that take only a string :
```groovy
Contract.make {
    description("""
Some description...
""")
    input 'the initial state of the provider'
    request {
       ...
    }
    response {
        ...
    }
}
```
[...]

PR #1743, opened 2021-12-14
Add capability to set initial request context. Fixes gh-1740.

mathieu-amblard:
Add context field for the request to be able to generate providerStates when a Contract for HTTP to a PACT.


marcingrzejszczak:
I saw that you've modified `Contract` to add `requestContext`. Maybe we can leverage the metadata section? That way for PACT we can add arbitrary things without impacting how Contract looks like. WDYT?

marcingrzejszczak:
The problem is that for the standard HTTP contract that uses MockMvc that field is irrelevant. I think we should go with the metadata approach.

mathieu-amblard:
Yes, you are right, so let's go for the metadata approach.

mathieu-amblard:
Hi @marcingrzejszczak,
I just push a new version with the metadata approach.
Thanks in advance for your feedbacks.

PR #1763, opened 2022-03-23
fix #1667: Add missing multipart support in Spock

### LINKED ISSUE #1667: Regression with generating multipart requests
After migration from Spring Cloud contract 2.2.7.RELEASE to 3.0.3. Cloud contract tests which use multipart content-type started failing with 400 Bad Request reason. After the analyse of the generated contracts I have found that one line is missed into the new version of Spring Cloud contracts.

Example of the contract - contract wasn't changed between the versions
```
Contract.make {
    request {
        method POST()
        url '/attachments'
        headers {
            header 'Content-Type': value(client(regex('multipart/form-data; ?boundary=.+')), server('multipart/form-data;boundary=AaB03x'))
        }
        multipart(
            file: named(
                name: $(c(regex(nonEmpty())), p('filename.csv')),
                content: $(c(regex(nonEmpty())), p('file content'))
            )
        )
    }
    response {
        status CREATED()
    }
}
```

It is almost the same as example provided here - https://cloud.spring.io/spring-cloud-contract/reference/html/project-features.html#contract-dsl-multiple

The generated code in Spring Cloud contract version 2.2.7.RELEASE
```
def validate_shouldReturnCreatedForUploadAttachments() throws Exception {
   given:
      RequestSpecification request = given()
            .header("Content-Type", "multipart/form-data;boundary=AaB03x")
            .multiPart('file', 'filename.csv', 'file content'.bytes)
   when:
      Response response = given().spec(request)
            .post("/attachments")
   then:
      response.statusCode() == 201
}
```

The generated code in  Spring Cloud contract version 3.0.3

```
	def validate_shouldReturnCreatedForUploadAttachments() throws Exception {
		given:
			RequestSpecification request = given()
					.header("Content-Type", "multipart/form-data;boundary=AaB03x")

		when:
			Response response = given().spec(request)
					.post("/attachments")

		then:
			response.statusCode() == 201
	}
```

It is possible to spot that the following line is missed
```.multiPart('file', 'filename.csv', 'file content'.bytes)```

When I have added this line manual and run the test with the new version of Spring Cloud contracts it passed however I am not able to force proper generation of the contract with this line.

igarash1:
Currently, the multipart support for EXPLICIT and WEBCLIENTTEST modes in Spock is missing. This PR adds the support.

Fixes #1667

### DIFF
--- a/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/builder/RestAssuredGiven.java
+++ b/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/builder/RestAssuredGiven.java
@@ -43,7 +43,9 @@ class RestAssuredGiven implements Given, BodyMethodVisitor, RestAssuredAcceptor 
 				.addAll(Arrays.asList(new MockMvcHeadersGiven(blockBuilder), new MockMvcCookiesGiven(blockBuilder),
 						new MockMvcBodyGiven(blockBuilder, generatedClassMetaData, bodyParser),
 						new JavaMultipartGiven(blockBuilder, generatedClassMetaData, bodyParser),
-						new SpockMockMvcMultipartGiven(blockBuilder, generatedClassMetaData, bodyParser)));
+						new SpockMockMvcMultipartGiven(blockBuilder, generatedClassMetaData, bodyParser),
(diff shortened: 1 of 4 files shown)


marcingrzejszczak:
Great work @igarash1 , thank you for the contribution!

PR #1783, opened 2022-05-13
Fixed relative URL computation in wIthoutBaseUrl

devdufutur:
Current behaviour :
base URL = http://foo.bar, url = http://foo.bar/my/api => my/api
base URL = http://foo.bar/, url = http://foo.bar/my/api => y/api

Fixed behaviour : 
base URL = http://foo.bar, url = http://foo.bar/my/api => /my/api
base URL = http://foo.bar/, url = http://foo.bar/my/api => my/api

### DIFF
--- a/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockRestServiceServer.java
+++ b/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockRestServiceServer.java
@@ -277,7 +277,7 @@ public final class WireMockRestServiceServer {
 		if (indexOfBaseUrl == -1) {
 			return url;
 		}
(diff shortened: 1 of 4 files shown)


marcingrzejszczak:
Can you create some tests for this?

devdufutur:
Sure, I did. Let me know if it's the way you wanted me to test. I would have prefer a unit test of withoutBaseUrl method in WireMockRestServiceServer but it's a private one :/

I modified existing test to reveal the false positive (it now fails without my fix and pass with it) and add two more to ensure matching by one character (one with a trailing slash in base URL and one with a  leading slash in urlPathPattern)

marcingrzejszczak:
Hmm sth is wrong with the build. Could you please file this PR against `3.1.x` branch ? I'll then merge this forward to `main`

devdufutur:
Of course. I did (https://github.com/spring-cloud/spring-cloud-contract/pull/1785) and closed this one.

PR #1785, opened 2022-05-17
Fixed relative URL computation in wIthoutBaseUrl

devdufutur:
Current behaviour :
* base URL = http://foo.bar/
* url = http://foo.bar/my/api
gives => my/api

Fixed behaviour : 
* base URL = http://foo.bar/
* url = http://foo.bar/my/api
gives => /my/api


marcingrzejszczak:
Great work @devdufutur

PR #1802, opened 2022-06-25
Give ability to place contracts in dedicated project

oleksiyp:
As a developer developing contracts
I would like to package tests into JAR
so that it is possible to execute in other project.

Currently running tests that are residing in JAR is not possible.
This is a fix that allow to read resources from JAR.
This is enough to make overall approach to work.


marcingrzejszczak:
Can you please create a test for this?

marcingrzejszczak:
Closing due to lack of feedback

Issue #1808, opened 2022-08-07
OptionalProperty produces wrong output with ClientDslProperty & ServerDslProperty


artemy:
## Bug description
`optional` around `ClientDslProperty`/`ServerDslProperty` properties such as `anyOf`, `anyNonBlankString`, `anyEmail`, etc. produce wrong output, like `OPTIONAL>>ClientDslProperty{\nclientValue=^FOO$|^BAR$, \n\tserverValue=FOO}<<`

**Version affected**: at least 3.1.3

## Sample
Sample repository: https://github.com/artemy/spring-cloud-contract-demo

<details>
<summary>Example contract</summary>

```kotlin
package contracts

import org.springframework.cloud.contract.spec.ContractDsl.Companion.contract

contract {
    request {
        method = GET
        url = url("/api")
        headers {
            contentType = APPLICATION_JSON
            accept = APPLICATION_JSON
        }
        body = body(
            "foo" to value(consumer(optional(anyOf("FOO", "BAR"))), producer("FOO")),
            "bar" to value(consumer(optional(anyNonBlankString)), producer("FOO")),
            "baz" to value(consumer(optional(nonBlank)), producer("BAZ"))
        )
    }
    response {
        status = OK
        headers {
            contentType = APPLICATION_JSON
        }
        body = body(
            "foo" to value(consumer("FOO"), producer(optional(anyOf("FOO", "BAR")))),
            "bar" to value(consumer("FOO"), producer(optional(anyNonBlankString))),
            "baz" to value(consumer("BAZ"), producer(optional(nonBlank)))
        )
    }
}
```
</details>

### Expected result:
<details>
<summary>generated test</summary>

```java
package com.github.artemy.springcloudcontractdemo;

import com.github.artemy.springcloudcontractdemo.ContractVerifierBaseIT;
import com.jayway.jsonpath.DocumentContext;
import com.jayway.jsonpath.JsonPath;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import io.restassured.module.mockmvc.specification.MockMvcRequestSpecification;
import io.restassured.response.ResponseOptions;

import static org.springframework.cloud.contract.verifier.assertion.SpringCloudContractAssertions.assertThat;
import static org.springframework.cloud.contract.verifier.util.ContractVerifierUtil.*;
import static com.toomuchcoding.jsonassert.JsonAssertion.assertThatJson;
[...]

PR #1809, opened 2022-08-07
Fix OptionalProperty producing wrong output with ClientDslProperty & ServerDslProperty

artemy:
Fix improperly accessed optional values; ServerDslProperty incorrectly wrapped by OptionalProperty

Fixes gh-1808


marcingrzejszczak:
Hey, thanks for this great PR. Can we do it against `3.1.x` branch? I will merge it forward to `main` then.

artemy:
I reckon you should be able to do it to `3.1.x` too, that's where I actually was developing and testing it

marcingrzejszczak:
OK so if this passes I'll cherry pick the commit to 3.1.x. No problem!

artemy:
There are some issues with the main branch pipeline

marcingrzejszczak:
Thanks for the PR ! Error is not related to your changes

PR #1815, opened 2022-08-30
fix: Fixed an issue that could not find a resources in the jar file

jhh992000:
I found a new bug while building the CI automation system.
When running the test in the cli environment, 
I found an issue that could not read the stubs file in the built jar file.
I changed it to read differently if stubs-uri's protocol is jar.


marcingrzejszczak:
Can you write a test that replicates this problem?

jhh992000:
I added a test case.
Please check.

marcingrzejszczak:
Fixed via https://github.com/spring-cloud/spring-cloud-contract/commit/bc6742b67e2dcebfd130d311c5f8c5c0b50a9276 , thanks for the PR!

PR #1870, opened 2023-02-18
Add child node's XML namespaces to Namespace Map

### LINKED ISSUE #1709: The problem with child node's XML namespaces
**Describe the bug**
I want to write tests in Kotlin for an SOAP service and there is a problem with child namespaces. Child node's namespaces break the application during build.

**Sample**
```
response {
body = body ("""
<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/">
        <soap:Body>
            <ns2:Res xmlns:ns2="http://*******/****/****/******/schema">
                <ns2:ID>1</ns2:ID>
            </ns2:Res>
        </soap:Body>
    </soap:Envelope>""")
}
```

**Possible problem**
Class:
org.springframework.cloud.contract.verifier.util.xml.DOMNamespaceContext

method:
```
private void addNamespaces (Node element) {
    if (element.getParentNode ()! = null) {
      addNamespaces (element.getParentNode ());
    }
    if (element instanceof Element) {
      Element el = (Element) element;
      NamedNodeMap map = el.getAttributes ();
      for (int x = 0; x <map.getLength (); x ++) {
        Attr attr = (Attr) map.item (x);
        if ("xmlns" .equals (attr.getPrefix ())) {
          namespaceMap.put (attr.getLocalName (), attr.getValue ());
        }
      }
    }
  }
```
As you can see, it only parses namespaces from the root (soap: Envelope) node. All namespaces declared in child nodes are ignored.

bensonbenny021:
Navigated through child nodes instead of parent and added namespace attributes if any to the namespace map

fixes #1709

### DIFF
--- a/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/util/xml/DOMNamespaceContext.java
+++ b/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/util/xml/DOMNamespaceContext.java
@@ -53,8 +53,13 @@ public class DOMNamespaceContext implements NamespaceContext {
 	}
 
 	private void addNamespaces(Node element) {
-		if (element.getParentNode() != null) {
-			addNamespaces(element.getParentNode());
+		if (element.getChildNodes() != null) {
+			// loops through child Element Nodes and check for namespace attributes
+			for (int i = 0; i < element.getChildNodes().length(); i ++) {
+				if (element.getChildNodes().item(i) instanceof Element) {
+					addNamespaces(element.getChildNodes().item(i));
+				}
+			}
 		}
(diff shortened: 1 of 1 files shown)


bensonbenny021:
Creating a PR here for the first time, Let me know if I have missed anything. Will try adding test cases for the same.

marcingrzejszczak:
At a first glance looks good, but we definitely need some tests for this

bensonbenny021:
@marcingrzejszczak Sorry I am facing issues while running the test cases in my local, I think some issues with my setup. Is there any recommendations on the setup?

marcingrzejszczak:
I don't know what issues you are facing

marcingrzejszczak:
Ping?

PR #1871, opened 2023-02-18
Attach up-to-date stubs to maven

### LINKED ISSUE #1860: spring-cloud-contract-maven-plugin: stubs are not attached as secondary artifact when contracts are unchanged
spring-cloud-contract-maven-plugin version 3.1.5

The stubs jar is not added as secondary artifact with classifier.
Calling mvn install twice results in first install with primary artifact + secondary stubs, the second only with primary.

The issue seems to be in GenerateStubsMojo#execute - Parameter incrementalContractStubsJar is true by default, thus stubs artifact is attached only when contracts change.
Setting incrementalContractStubsJar to false "fixes" it.

The workflow is something like this:
- mvn package (stubs jar present)
- Do some other checks
- Wait for deploy confirmation
- mvn deploy (stubs jar missing)

NewAgeCZ:
Fixes #1860 

* Add failsafe for integration tests
* Move currently ignored PluginUnitTest into integration test phase
* Fix reported issue and add test for it.

The surefire/failsafe plugin configuration feels a bit raw to me, yet sufficient for now. What do you think?


marcingrzejszczak (approved):
LGTM - I'll cherry-pick that for branch 3.1.x.

Issue #1818, opened 2022-09-20
Support priority in Restdocs generated contract


huberchrigu:
**Is your feature request related to a problem? Please describe.**
I use Spring REST Docs to generate dslContract() snippets. I have multiple contracts with similar request parameters, so I need to specify the priority in order to fix my tests on consumer side.
I tried _dslContract(mapOf("priority" to 1))_, but it does not work.

**Describe the solution you'd like**
Either add an example in the documentation how to do it or extend the snippet with an optional priority attribute.

PR #1882, opened 2023-04-05
Adds support to specify priority and allow for template override

wwadge:
Allows specifying priority eg

```
SpringCloudContractRestDocs.dslContract(Maps.of("priority", 1))))
```

and makes it possible to override the dsl template
```
 new ContractDslSnippet(){
                            @Override
                            protected String getTemplate() {
                                return "custom-dsl-contract";
                            }
                        }));
```

(template must live in resources/org/springframework/restdocs/templates)

Fixes gh-1818


marcingrzejszczak:
Thanks @wwadge for your PR!  Actually one more question, can you also add a PR with docs update that will mention this feature?

wwadge:
Sure - let me see if I can find the right docs to update and open a new PR

Issue #1911, opened 2023-06-26
Gradle 8.1.1 task declaration problems


marcingrzejszczak:
(3.1.x contract version) after having fixed the deprecation issues I get this one

```groovy
* What went wrong:
A problem was found with the configuration of task ':copySnippets' (type 'Copy').
  - Gradle detected a problem with the following location: '/home/marcin/repo/spring-cloud-contract/samples/standalone/webclient/http-server/build/stubs/META-INF/com.example/http-server-webclient-gradle/0.0.1/mappings'.
    
    Reason: Task ':verifierStubsJar' uses this output of task ':copySnippets' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':copySnippets' as an input of ':verifierStubsJar'.
      2. Declare an explicit dependency on ':copySnippets' from ':verifierStubsJar' using Task#dependsOn.
      3. Declare an explicit dependency on ':copySnippets' from ':verifierStubsJar' using Task#mustRunAfter.
    
    Please refer to https://docs.gradle.org/8.1.1/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```

@shanman190 :grimacing:

marcingrzejszczak:
Fixed by https://github.com/spring-cloud/spring-cloud-contract/commit/5c6eac2453ac1761c9f8e1a1ac1c41d6e55b9f8d

shanman190:
This one is pretty easy, so I'll open a PR for the actual fix to 3.1.x sometime today.

marcingrzejszczak:
Thank you!

PR #1913, opened 2023-06-26
Add dependency on copyContracts

shanman190:
Fixes gh-1911.

### DIFF
--- a/samples/standalone/dsl/http-server/build.gradle
+++ b/samples/standalone/dsl/http-server/build.gradle
@@ -45,9 +45,9 @@ dependencies {
 contractTest {
 	useJUnitPlatform()
 	systemProperty 'spring.profiles.active', 'gradle'
-	// testLogging {
-	// 	exceptionFormat = 'full'
-	// }
+	testLogging {
+		exceptionFormat = 'full'
+	}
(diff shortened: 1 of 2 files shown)


shanman190:
I kept the same style that the project was using for adding the dependency. Normally this would have looked more like:

```
TaskProvider<Jar> verifierStubsJar = project.getTasks().register(VERIFIER_STUBS_JAR_TASK_NAME, Jar.class);
verifierStubsJar.configure(stubsJar -> {
	stubsJar.setDescription("Creates the stubs JAR task");
	stubsJar.setGroup(GROUP_NAME);
	stubsJar.getArchiveBaseName().convention(project.provider(project::getName));
	stubsJar.getArchiveClassifier().convention(extension.getStubsSuffix());
	stubsJar.from(copyContracts, generateClientStubs);
});
```

The other aspect of why I kept it this way was that each task is reusing the same output directories as well which makes the ability to cache these tasks correctly a bit more difficult, but it still results in the same result. In order to fix that, it would mean having output directories that are unique to each task, then taking those outputs as inputs into subsequent tasks. That was a lot more change than I thought was necessary, so the simpler solution was used.

Issue #1888, opened 2023-04-17
Provide the ability to specify the Content Type of a non-named Multipart Part


adrianhj:
**Is your feature request related to a problem? Please describe.**

When utilising @RequestPart on a Controller argument in Spring for working with multipart/form-data requests consisting of e.g. a file part and a JSON part—and expecting the JSON part to be correctly deserialised to a POJO—it relies on the Content-Type of the request part to find and apply the relevant `HttpMessagingConverter`; defaulting to `application/octet-stream` if none was provided.

From the client perspective these parts are not files as such and do not have file names. [MultipartBodyBuilder.part(String name, Object part, MediaType contentType)](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/client/MultipartBodyBuilder.html#part(java.lang.String,java.lang.Object,org.springframework.http.MediaType)) for example allows specifying such a part with a relevant content type when working with the WebClient.

**Describe the solution you'd like**

Provide the ability to specify the content type of a multipart part within a contract so as to generate valid requests.

**Describe alternatives you've considered**

Utilising the `named(name, content, contentType)` method, however this puts the burden onto any integrating party to provide a dummy name to satisfy the generated WireMock mappings as it is not optional.

**Additional context**

REST Assured seems to have various flavours of `multiPart`, some providing the ability to specify a multipart part with a relevant content type as desired which is probably of interest as compared to the currently used `param` method which seems more intended for form params.

In the process of attempting to provide a `part(content, contentType)` method or the like, but figured I'd open for discussion if this is something that has been covered previously or whether I have misunderstood any aspect of verifying a JSON encoded @RequestPart via Spring Cloud Contract.

marcingrzejszczak:
Do you have a small Java-Maven sample that you could show with such client - server communication?

adrianhj:
@marcingrzejszczak I have built a small example in https://github.com/adrianhj/spring-cloud-contract-1888, hopefully this helps showcase the gap (issue?) utilising Spring.

Areas of interest:

* The Controller being called: https://github.com/adrianhj/spring-cloud-contract-1888/blob/main/src/main/java/com/example/springcloudcontract1888/SpringCloudContract1888Application.java#L24
* A test showing the endpoint being called via RestTemplate: https://github.com/adrianhj/spring-cloud-contract-1888/blob/main/src/test/java/com/example/springcloudcontract1888/SpringCloudContract1888ApplicationTests.java#L33
[...]

PR #1929, opened 2023-08-17
feat: allow multipart parts with content types

adrianhj:
WIP implementation for commentary of an alternative approach for `multipart/form-data` support based off of the back of #1888 backed by WireMock's multipart support vs generating a regex against the request parts.

Looking to pick this back up and progress if there is interest with some guidance.

DSL wise aim would be to provide something closer to the `multipart/form-data` spec to give a bit more freedom than currently possible, such as the following for a scenario of sending a correctly structured JSON metadata part and a file part in the request.

```
multipart(
    metadata: part(
        value: "{...}",
        contentType: "application/json",
    ),
    file: part(
        value: <file>,
        filename: "example.pdf",
        contentType: "application/pdf"
    )
)
```


marcingrzejszczak:
Also since this is a new feature I think we should add it to `main`, WDYT ?

PR #1996, opened 2023-09-26
Use prepareTestInstance phase instead of beforeTestClass phase

### LINKED ISSUE #1995: WireMockTestExecutionListener blocks any other before test actions
**Describe the bug**
The `WireMockTestExecutionListener` in the `beforeTestClass` checks whether the application context is not broken by calling `testContext.getApplicationContext();`, which blocks all other actions until the application context is initialised. In turn this makes any other before actions (ie. BeforeAll annotated methods) to be run after the whole application context is initialized and it may be to late for these methods (ie. test container setup methods). This is called even when the test class doesn't have any wiremock configuration setup.
 So probably it could be improved by having a different if condition order - `if (annotationMissing(testContext) || wireMockConfigurationMissing(testContext) || applicationContextBroken(testContext))` and by checking whether the context is actually available via `testContext.hasApplicationContext` in the `applicationContextBroken` method, as there shouldn't be a need to see if it's broken when it's not there yet?

Fadelis:
Closes #1995 
In the end I choose to use a different test execution phase `prepareTestInstance` which is called after `applicationContext` is available, since that whole method depends on it being available.

### DIFF
--- a/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockTestExecutionListener.java
+++ b/spring-cloud-contract-wiremock/src/main/java/org/springframework/cloud/contract/wiremock/WireMockTestExecutionListener.java
@@ -39,7 +39,7 @@ public final class WireMockTestExecutionListener extends AbstractTestExecutionLi
 	private static final Log log = LogFactory.getLog(WireMockTestExecutionListener.class);
 
 	@Override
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
@Fadelis there are some checkstyle issues. Here you have hints on contributing https://github.com/spring-cloud/spring-cloud-build/blob/main/docs/modules/ROOT/partials/contributing.adoc

marcingrzejszczak (approved):

Issue #2035, opened 2023-10-31
AutoConfigureWiremock with a pre-defined custom port always defaults to port 8080


igdianov:
**Describe the bug**

AutoConfigureWiremock with a pre-defined custom port always defaults to port 8080 because of this change: https://github.com/spring-cloud/spring-cloud-contract/commit/a15db490f169412e50fc3140d95d8a70108465d5

Spring Cloud Version: 4.1.0-M2

**Sample**

```
@SpringBootTest
@AutoConfigureWireMock(port = 8888)
public class WiremockImportApplicationTests {

	@Autowired
	private WireMockServer wireMockServer;

	@Test
	public void contextLoads() throws Exception {
		assertThat(this.wireMockServer.port()).isEqualTo(8888);
	}

}
```

```
org.opentest4j.AssertionFailedError: 
expected: 8888
 but was: 8080
Expected :8888
Actual   :8080
<Click to see difference>


	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at com.alfresco.form.runtime.controller.WiremockImportApplicationTests.contextLoads(WiremockImportApplicationTests.java:23)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```

igdianov:
@spencergibb Any chance this gets triaged and resolved in the upcoming Spring Boot 3.2 release?

spencergibb:
It's not part of spring boot 3.2, but spring cloud 2023.0, I think it can.

igdianov:
@spencergibb Any chance this gets resolved soon for 2023.0 release?

igdianov:
Any updates on this, please? The PR is ready. Just needs 👍

andrea-ligios:
Hi @spencergibb , this is a low hanging fruit 👍🏻  

[...]

PR #2036, opened 2023-10-31
Fix GH-2035 AutoConfigureWiremock with a pre-defined custom port always defaults to port 8080

igdianov:
Fixes https://github.com/spring-cloud/spring-cloud-contract/issues/2035


marcingrzejszczak:
Thanks @igdianov for your contribution!

PR #2077, opened 2023-12-28
Update pom.xml

jawscout:
Update Guava version to 33.0.0-jre


marcingrzejszczak:
I need to revert this cause we'll need to do it once main points to 4.2.0

PR #2097, opened 2024-02-02
Revert "Bumping versions"

cameleeck:
This reverts commit e47f7f89e63b579ae7bae26503bd7014965c02ae.


spencergibb:
I don't think we want to do this.

cameleeck:
@spencergibb what you propose? It seems that it bumped version backwards and right now we have 4.1.1-SNAPSHOT and 4.1.2-SNAPSHOT -> manually update everything to 4.1.2-SNAP?

spencergibb:
We need to fix things on our end

PR #2156, opened 2025-02-17
fix: Use exec for better PID 1 handling

jakub-bochenski:
Also pass the arguments along to the java execution

Previously the arguments were silently dropped

### DIFF
--- a/docker/spring-cloud-contract-stub-runner-docker/run.sh
+++ b/docker/spring-cloud-contract-stub-runner-docker/run.sh
@@ -14,5 +14,9 @@ if [[ "${MESSAGING_TYPE}" != "" ]]; then
   ADDITIONAL_OPTS="${ADDITIONAL_OPTS} --thin.profile=${MESSAGING_TYPE}"
 fi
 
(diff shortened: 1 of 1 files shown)


jakub-bochenski:
> Failed to execute goal org.antora:antora-maven-plugin:1.0.0-alpha.5:antora (default) on project spring-cloud-contract-docs: Antora failed to generate site successfully: Unable to execute com.github.eirslett:frontend-maven-plugin:1.15.0:npx (inspect prior log messages to find cause): Failed to run task: 'npx --offline antora --to-dir target/antora/site --stacktrace --fetch antora-playbook.yml' failed. org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1) -> [Help 1]

I don't think this PR broke that

marcingrzejszczak:
I am no expert with Docker - I didn't know about that problem, thanks for pointing it out! 

Yeah, don't worry about the issue with antora. Can you check the DCO check please?

marcingrzejszczak:
Yeah, the antora build error is not related to your changes

PR #2309, opened 2025-10-01
stubrunner. -> spring.cloud.contract.stubrunner. property migration

marcingrzejszczak:


timtebeek:
hi @marcingrzejszczak ; hope you're well! Elsewhere for property migrations we've seen deprecation entries added as seen here:
```json
    {
      "name": "logging.file.clean-history-on-start",
      "type": "java.lang.Boolean",
      "description": "Whether to clean the archive log files on startup. Only supported with the default logback setup.",
      "sourceType": "org.springframework.boot.context.logging.LoggingApplicationListener",
      "defaultValue": false,
      "deprecation": {
        "replacement": "logging.logback.rollingpolicy.clean-history-on-start",
        "since": "2.4.0"
      }
    },
```
Would it be possible to back fill those in https://github.com/spring-cloud/spring-cloud-contract/blob/v5.0.1/spring-cloud-contract-stub-runner/src/main/resources/META-INF/additional-spring-configuration-metadata.json ?

That way Spring Tools users will get notified about the suggested replacements, as well as other tools using that metadata. 😇

marcingrzejszczak:
There you go https://github.com/spring-cloud/spring-cloud-contract/pull/2398

PR #2327, opened 2025-10-26
Changes For Boot 4.0.0 RC1 Compatibility

ryanjbaxter:


marcingrzejszczak:
:shipit:

ryanjbaxter:
@marcingrzejszczak not sure what that means 😁 but at least the build passes.  I still have some more investigation to do though

spring-cloud-contract-stub-runner/src/test/groovy/org/springframework/cloud/contract/stubrunner/AetherStubDownloaderSpec.groovy
```diff
@@ -29,9 +28,6 @@ import org.springframework.util.ResourceUtils
 
 class AetherStubDownloaderSpec extends Specification {
 
-	@Rule
```
ryanjbaxter: I don't really know why this rule was here in the first place, we were not using Hoverfly in any of the tests from what I can tell

PR #2367, opened 2025-12-03
fix: Adds an empty list to nonFilteredFileExtensions

### LINKED ISSUE #2361: Maven plugin fails when converting resources
**Describe the bug**
Hello! 
Trying to upgrade to the new major release (Spring Boot 4, Spring Cloud 2025.1 and Spring Cloud Contract 5), I encounter an issue while trying to run my contract tests. When I execute `mvn clean install` (or more precisely `mvn clean org.springframework.cloud:spring-cloud-contract-maven-plugin:5.0.0:convert`), I get the following error:  

```
[ERROR] Failed to execute goal org.springframework.cloud:spring-cloud-contract-maven-plugin:5.0.0:convert (default-convert) on project beer-api-producer: Execution default-convert of goal org.springframework.cloud:spring-cloud-contract-maven-plugin:5.0.0:convert failed: Cannot invoke "java.util.List.contains(Object)" because the return value of "org.apache.maven.shared.filtering.MavenResourcesExecution.getNonFilteredFileExtensions()" is null -> [Help 1]
```

Note that removing the `clean` goal and re-running the `convert` command a second time makes it work.

**Sample**  
Can be reproduced using your sample here :  
https://github.com/spring-cloud-samples/spring-cloud-contract-samples/tree/main/producer

marcingrzejszczak:
without this change an NPE is being thrown from Maven Resources Filtering with this change since the list is empty an NPE is no longer thrown

fixes #2361

### DIFF
--- a/spring-cloud-contract-tools/spring-cloud-contract-maven-plugin/src/main/java/org/springframework/cloud/contract/maven/verifier/CopyContracts.java
+++ b/spring-cloud-contract-tools/spring-cloud-contract-maven-plugin/src/main/java/org/springframework/cloud/contract/maven/verifier/CopyContracts.java
@@ -82,7 +82,8 @@ class CopyContracts {
 		execution.setOverwrite(true);
 		execution.setIncludeEmptyDirs(false);
 		execution.setFilterFilenames(false);
-		execution.setFilters(Collections.emptyList());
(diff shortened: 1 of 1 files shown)


jduhking:
what version is this fix in now? Is it released yet?

PR #2398, opened 2026-01-06
Added migration for deprecated properties

marcingrzejszczak:
related to gh-2309


timtebeek:
Thanks a lot for the quick work here! Should make it easier for folks to find the replacements to adapt to the earlier changes.

spencergibb (approved):

PR #2405, opened 2026-01-17
ContractVerifierObjectMapper - special support for Avro

### LINKED ISSUE #2402: ContractVerifierObjectMapper fails for Avro-generated objects
**Describe the bug**
The `ContractVerifierObjectMapper` maps the actual intercepted message to JSON, before performing the relevant assertions on it. 

However, if the message is an Avro-generated object, mapping it to JSON will fail due to some of the avro-specific fields in that record (schema, specificData, classSchema, conversion)

**Sample**
I have reproduced the issue [here](https://github.com/spring-cloud/spring-cloud-contract/compare/main...etrandafir93:spring-cloud-contract:contracte_verifier_avro). 
(For simplicity, I have just copied and pasted an Avro-generated class)

In my project, I have configured the ContractVerifierObjectMapper's JsonMapper to ignore these problematic fields, and it solved the issue. 

Would it be fair to assume that, if Avro is in the classpath, we should ignore these 4 fields while converting to JSON? 
If so, we can encapsulate this bit of logic inside the `ContractVerifierObjectMapper`

etrandafir93:
Some Avro-specific fields will make the ContractVerifierObjectMapper when it tries to map the actual intercepted message to JSON, before performing the relevant assertions on it.

Ignoring these fields will fix the issue, and it won't affect the JSON assertions.

closes #2402


spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/noop/NoOpContractVerifierAutoConfiguration.java
```diff
@@ -89,12 +92,27 @@ public ContractVerifierMessaging<Object> contractVerifierMessaging() {
...
-		}
-		return new ContractVerifierObjectMapper();
+	@ConditionalOnMissingClass("org.apache.avro.specific.SpecificRecordBase")
+	public ContractVerifierObjectMapper contractVerifierObjectMapper(
+			ObjectProvider<JsonMapper> jsonMapper) {
+		JsonMapper mapper = jsonMapper.getIfAvailable(JsonMapper::new);
+		return new ContractVerifierObjectMapper(mapper);
+	}
+
+	@Bean
```
marcingrzejszczak: I would wrap these Avro related classes in a nested `AvroContractVerifierObjectMapperConfiguration`

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/noop/NoOpContractVerifierAutoConfiguration.java
```diff
@@ -99,20 +99,25 @@ public ContractVerifierObjectMapper contractVerifierObjectMapper(
 		return new ContractVerifierObjectMapper(mapper);
 	}
 
-	@Bean
-	@ConditionalOnMissingBean
+	@Configuration(proxyBeanMethods = false)
 	@ConditionalOnClass(name = "org.apache.avro.specific.SpecificRecordBase")
```
marcingrzejszczak: Since it's a cass you can add this as an optional dependency and then normally reference it. Then same below - you could use the class reference instead of doing Class.forName()

marcingrzejszczak (approved):

PR #2408, opened 2026-01-19
Remove commons-beanutils usage

### LINKED ISSUE #2129: Get rid of commons-beanutils
Spring Cloud Contract uses `commons-beanutils` which is extremely old and depends on even more outdated and moreover vulnerable `commons-collections` library, there is also no evidence that `common-beanutils` will be updated any time soon.

The use of the aforementioned library is quite limited and should be easy to rewrite without use of this library.
I actually found only one usage: https://github.com/spring-cloud/spring-cloud-contract/blob/d771b76c333508a85e1bc672f46b0b5ab2dae1f4/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/builder/JsonBodyVerificationBuilder.java#L303

Also there are several usages of `commons-collections` in one file:
https://github.com/spring-cloud/spring-cloud-contract/blob/d771b76c333508a85e1bc672f46b0b5ab2dae1f4/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/converter/YamlToContracts.java#L43
Which can be replaced with Spring specific implementations or `commons-collections4` if it's not possible.

marcingrzejszczak:
fixes #2129

### DIFF
--- a/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/builder/BodyAssertionLineCreator.java
+++ b/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/builder/BodyAssertionLineCreator.java
@@ -70,6 +70,9 @@ class BodyAssertionLineCreator {
 			return getResponseBodyPropertyComparisonString(singleContractMetadata, property,
 					((DslProperty) value).getServerValue());
 		}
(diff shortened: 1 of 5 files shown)


ryanjbaxter:
Sounds good to me

ryanjbaxter:
Oh I believe it 🤦‍♂️

PR #2468, opened 2026-05-06
Avro support

### LINKED ISSUE #2402: ContractVerifierObjectMapper fails for Avro-generated objects

### LINKED ISSUE #2404: StubRunnerExecutor fails for Avro objects
**Describe the bug**
[StubRunnerExecutor::sendMessage](https://github.com/spring-cloud/spring-cloud-contract/blob/main/spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/StubRunnerExecutor.java#L245) converts the message to a JSON string before delegating to the `MessageVerifierSender` that should publish it:

```java
void sendMessage(Contract groovyDsl) {
    // ...	

    Object payload = null;
    if (body != null && body.getClientValue() instanceof FromFileProperty) {
        FromFileProperty fromFile = (FromFileProperty) body.getClientValue();
        if (fromFile.isByte()) {
            payload = fromFile.asBytes();
        }
        else {
            payload = fromFile.asString();
        }
    }
    else {
        payload = JsonOutput
            .toJson(BodyExtractor.extractClientValueFromBody(body == null ? null : body.getClientValue()));
            // here! Even if "body" is an Avro object, it will be converted to json
    }

    this.messageVerifierSender.send(payload, headers == null ? null : headers.asStubSideMap(),
        outputMessage.getSentTo().getClientValue(), contract);
}
```

The [MessageVerifierSender](https://github.com/spring-cloud/spring-cloud-contract/blob/main/spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/MessageVerifierSender.java) uses generic types, and it doesn't mention that the message comes as a JSON String.

When implementing the `MessageVerifierSender::send`, we'll most likely use an API for externalizing the message _(eg: KafkaTemplate for Kafka)_. In this case, KafkaTemplate already encapsulates the logic for serializing the payload (as per the configured KafkaProperties). 

In other words, the KafkaTemplate from _MessageVerifierSender::send_ tries to serialize the payload as Avro, but it fails to do so (or ends up with a weird result) because the object passed to the _MessageVerifierSender_ is a JSON string. 

---

**Ugly workaround:** 
When implementing the _MessageVerifierSender_, I had to deserialize the JSON back to my Avro object, and (only then) let KafkaTemplate do the serialization.

etrandafir93:
- **Bug fix #2402 — `ContractVerifierObjectMapper` fails for Avro objects**: when the intercepted message is an Avro-generated object, `ContractVerifierObjectMapper` would fail to serialize it to JSON due to Avro-specific fields (`schema`,   
  `specificData`, `classSchema`, `conversion`). Fixed by configuring the `JsonMapper` to ignore those fields via a mixin when Avro is on the classpath.                                                                                             
                                                                                                                                                                                                                                                    
  - **Avro support for contract-based messaging**: introduced `KafkaAvroMessageVerifierSender` which builds an Avro `GenericRecord` from the contract body and sends it via `KafkaTemplate`, backed by `KafkaAvroContractVerifierConfiguration`     
  (auto-configuration) and `AvroMetadata` (schema config in contract metadata).                                                                                                                                                                     

  - **Header propagation**: `KafkaAvroMessageVerifierSender` now wraps the payload in a `ProducerRecord` and copies contract output headers (e.g. `X-Correlation-Id`) as UTF-8 bytes onto it, so they are actually sent to Kafka.

  - **Bug fix #2404 — `StubRunnerExecutor` JSON-serializes Avro body**: `StubRunnerExecutor.sendMessage()` was unconditionally calling `JsonOutput.toJson()` on the message body before passing it to the `MessageVerifierSender`. This broke
  `KafkaAvroMessageVerifierSender` which expects a `Map`, not a JSON string. Fixed by adding an `isAvroContract()` check that inspects the raw contract metadata and skips JSON serialization for Avro contracts, passing the body as a `Map`
  directly.

TODO: if this gets accepted, we also need to add the documentation and sthe samples repo

  ## Related issues
  - Fixes #2402
  - Fixes #2404


spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/StubRunnerExecutor.java
```diff
@@ -269,6 +273,25 @@ private void sendMessage(Contract groovyDsl) {
 				outputMessage.getSentTo().getClientValue(), contract);
 	}
 
+	private boolean isFromFileProperty(DslProperty<?> body) {
+		return body != null && body.getClientValue() instanceof FromFileProperty;
+	}
+
+	private boolean isAvroContract(YamlContract contract) {
```
marcingrzejszczak: Can we make this section kafka / avro agnostic? If we just set asBytes() it should work right? Another option is to verify the contentType in which case if it's not json but explicitly sth else then we just pass through? Wdyt?

I also sense that we could have some extension points here. Like check through spi (we already do it in other parts of scc) some interfaces to verify if this contract has a special way of treating payload. We could have some AvroContractPayloadProceesor that would activate when the metadata has avro and it would provide a function to how convert the payload. That way this core logic stays clean of avro but we can inject the behavior. There would have to be some priority / ordering like we do in other spis here in scc.

Wdyt?

spring-cloud-contract-stub-runner/src/test/groovy/org/springframework/cloud/contract/stubrunner/StubRunnerExecutorAvroSpec.groovy
```diff
@@ -0,0 +1,87 @@
...
+	def 'should send Avro-serialized GenericRecord to Kafka for Avro contracts (bug #2404)'() {
+		given:
+			def tmpContractDir = saveTmpContract("""
+label: book_returned
+input:
+  triggeredBy: publishBookReturned()
+outputMessage:
+  sentTo: book.returned
+  headers:
+    X-Correlation-Id: abc-123-def
```
marcingrzejszczak, line 39: What headers are being sent in case of avro messages? Is there any content type?

spring-cloud-contract-stub-runner/src/test/groovy/org/springframework/cloud/contract/stubrunner/StubRunnerExecutorAvroSpec.groovy
```diff
@@ -0,0 +1,87 @@
...
+				header(it, "X-Correlation-Id") == "abc-123-def"
+			})
+		cleanup:
+			executor.shutdown()
+			tmpContractDir.deleteDir()
+	}
+
+	private File saveTmpContract(String contractYaml) {
+		File contractDir = File.createTempDir()
+		new File(contractDir, "book_returned.yml").text = contractYaml
```
marcingrzejszczak, line 79: Add a return

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/avro/KafkaAvroContractVerifierConfiguration.java
```diff
@@ -0,0 +1,85 @@
...
+		JsonMapper mapper = jsonMapper.getIfAvailable(JsonMapper::new)
+			.rebuild()
+			.addMixIn(SpecificRecordBase.class, IgnoreAvroMixin.class)
+			.build();
+		return new ContractVerifierObjectMapper(mapper);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(name = "avroKafkaTemplate")
+	KafkaTemplate<String, Object> avroKafkaTemplate(@Value("${spring.kafka.bootstrap-servers}") String bootstrapServers,
```
marcingrzejszczak: Why can't we reuse the users production template?

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/avro/KafkaAvroContractVerifierConfiguration.java
```diff
@@ -0,0 +1,85 @@
...
+
+	@Bean
+	@ConditionalOnMissingBean
+	KafkaAvroMessageVerifierSender kafkaAvroMessageVerifierSender(
+			@Qualifier("avroKafkaTemplate") KafkaTemplate<String, Object> avroKafkaTemplate) {
+		return new KafkaAvroMessageVerifierSender(avroKafkaTemplate);
+	}
+
+	@JsonIgnoreProperties({ "schema", "specificData", "classSchema", "conversion" })
+	interface IgnoreAvroMixin {
```
marcingrzejszczak: Can you provide a javadoc why we need this?

spring-cloud-contract-stub-runner/src/test/groovy/org/springframework/cloud/contract/stubrunner/StubRunnerExecutorAvroSpec.groovy
etrandafir93: afaik, there are no headers strictly required by the kafka broker or the Avro/confluent deserializer — the schema info is embedded in the payload. `contentType` is sometimes added by convention but isn't enforced by anything.                                

This `X-Correlation-Id` header in the test is just a custom header used to verify that user-defined headers are propagated correctly. Would you prefer to rename it to something more obviously test-scoped (e.g. `X-Dummy-Custom-Header`), or add a `contentType: application/avro` ?

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/avro/KafkaAvroContractVerifierConfiguration.java
etrandafir93: I was thinking about it: this is meant to be used on the consumer side - therefore we can have a kafkaTemplate configured differently here (eg: seems unlikely, but we can have an app consuming avro and publishing json)

What do you think about looking for a bean named avroKafkaTemplate first, if not, fall back to the "prod" kafkaTempalte bean, if not present either, create one.

```java
 if (ctx.containsBean("avroKafkaTemplate")) {
  return (KafkaTemplate<String, Object>) ctx.getBean("avroKafkaTemplate");
}
if (ctx.containsBean("kafkaTemplate")) {
  return (KafkaTemplate<String, Object>) ctx.getBean("kafkaTemplate");
}
return createAvroKafkaTemplate(bootstrapServers, schemaRegistryUrl);
``` 

My main point is to allow overriding this `avroKafkaTemplate` used by the stub without altering the prod bean.
What do you think about smth like this?

spring-cloud-contract-stub-runner/src/main/java/org/springframework/cloud/contract/stubrunner/StubRunnerExecutor.java
etrandafir93: > Can wIf we just set asBytes() it should work right?

by passing bytes, do you mean the 2nd approach demo-ed [here](https://github.com/spring-cloud-samples/spring-cloud-contract-samples/pull/429)?

If so - yes, it works, but it becomes very hard to peer review the contract changes and understand test failures. 

In fact, if we stick to that approach, we won't need any of this - as it mainly covers workarounds needed for the 1st apporach described in the samples PR.

> Another option is to verify the contentType in which case if it's not json but explicitly sth else then we just pass through? Wdyt?

Are you suggesting setting a header like `contentType: application/avro` and relying on it here, instead of the new custom field we added to the contract metadata? 

It would force users to add the _contentType_ header to the contract even if they don't use it in prod (as it is not strictly required, more like a convention **some** are using).  

This would work - but imo sounds a bit sketchy. wdyt? 
What's your concern about using the contract meta here? 

> I also sense that we could have some extension points here

Yes, that's a great idea, actually. We have a list of processors, each of them will provide a predicate to test the contract (therefore, we see if we should apply it) , and a function for mapping/extracting the payload. 

This way, we can keep using the metadata approach, but extract it somewhere else and keep this agnostic of avro. 

+You mentioned potentially adding prtobuff later, which might use this extension point too.

spring-cloud-contract-stub-runner/src/test/groovy/org/springframework/cloud/contract/stubrunner/StubRunnerExecutorAvroSpec.groovy
etrandafir93: done

spring-cloud-contract-verifier/src/main/java/org/springframework/cloud/contract/verifier/messaging/avro/KafkaAvroContractVerifierConfiguration.java
etrandafir93: done

PR #2475, opened 2026-05-18
Pin Kotlin version

dsyer:
Will fix compatibility builds with Spring Boot 4.1.0
(but requires users of Spring Boot 4.1.0 to do the same
so needs some documentation).


sdeleuze:
I propose https://github.com/spring-cloud/spring-cloud-contract/pull/2476 instead so IMO this PR should be closed unmerged.

